PR #22, opened 2014-08-15
Use the same ObjectId generation algorithm in native code

ConradIrwin:
Without the call to htonl() object ids generated within the same second have an unusually high probability of being in the "wrong order", which caused problems for the Bugsnag test-suite when upgrading to mongoid 4.


durran:
Merged into master.

PR #25, opened 2014-09-19
A shortcut to BSON::ObjectId.from_string

ronanrodrigo:
A shortcut to BSON::ObjectId.from_string('048fec556265664f5b110000') instead this, just use '048fec556265664f5b110000'.to_object_id


durran:
Thanks for the pull request, but it's better to keep the enhancements on Ruby's core string to a minimal and will not pull this in.

PR #26, opened 2014-10-02
Implements BSON::Binary#inspect.

haihappen:
With the current implementation of `inspect` writing a lot of
binary data results in *huge* log files. This commit fixes the
problem by returning only `BSON::Binary:"object_id"`.

### DIFF
--- a/lib/bson/binary.rb
+++ b/lib/bson/binary.rb
@@ -96,6 +96,18 @@ module BSON
       @type = type
     end
 
+    # Get a nice string for use with object inspection.
+    #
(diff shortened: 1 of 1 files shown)


huanzhang:
@durran could you tell me what's your email address, as I want to ask you some questions about generating ObjectId with BSON::ObjectId.

durran:
Merged in, added type and truncated data to 8 chars just to give a little bit of information... Thanks.

haihappen:
Thanks too!

PR #30, opened 2015-04-22
RUBY-898 Adjust Boolean#from_bson for IO streams that return String from #readbyte

estolfo:


lib/bson/boolean.rb
```diff
@@ -37,7 +37,7 @@ class Boolean
     #
     # @since 2.0.0
     def self.from_bson(bson)
-      bson.readbyte == 1
+      bson.readbyte.chr.force_encoding('BINARY') == TrueClass::TRUE_BYTE
```
durran: What caused the need to `force_encoding` here? We should always be getting binary strings on reads.

durran:
Looks good. Liked the way you tested it as well.

estolfo:
It was a copy and paste from here https://github.com/mongodb/bson-ruby/blob/master/lib/bson/boolean.rb#L28
But you're right, we don't need to encode to BINARY when reading.

PR #33, opened 2015-06-04
Issue when using a binary field with mongoid

keysen:
After your recent changes we are not able to use a binary field with mongoid.

We used ActiveSupport::Gzip to compress a string, then we have a binary encoded to ascii-8bit, which is the default encoding for a binary, but when mongoid call your ".inspect" method, it raises because incompatible encoding type. It was expected to get utf8 encoding but receives ascii8-bit.

### DIFF
--- a/lib/bson/binary.rb
+++ b/lib/bson/binary.rb
@@ -117,7 +117,7 @@ module BSON
     #
     # @since 2.3.0
     def inspect
-      "<BSON::Binary:0x#{object_id} type=#{type} data=#{data[0, 8]}...>"
+      "<BSON::Binary:0x#{object_id} type=#{type} data=#{data[0, 8].unpack('H*')}...>"
     end
 
     # Encode the binary type
--- a/lib/bson/binary.rb
+++ b/lib/bson/binary.rb
@@ -117,7 +117,7 @@ module BSON
     #
     # @since 2.3.0
     def inspect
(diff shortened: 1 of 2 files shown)


jarthod:
Indeed that's an issue, but your fix to display hex instead looks crappy with the array inspect, you should simply return the hex string, maybe prefixed with "0x"

keysen:
Ok, it's a bit better :)

gevans:
:+1: This is the only issue I'm running into after upgrading for the recent fix in 976da329ff03ecdfca3030eb6efe3c85e6db9999.

elhu:
Thank you very much @durran ! 
Any chance to get a new version of the gem release, as to avoid using the repo directly?

adrianmacneil:
I just ran into this as well, after upgrading to moped v2.0.5 (and therefore bson v3.0.4).

Since I couldn't find anything via Google, here is the actual error message:

    Encoding::CompatibilityError: inspected result must be ASCII only or use the default external encoding

Unfortunately moped targets `bson ~> 3.0.4`, so if you are planning to release the next version of bson gem as v3.1.0, is it possible to release a new moped version too with updated bson requirement?

durran:
BSON 3.1.0 is out, and I released Moped 2.0.6 with a slightly relaxed dependency `bson ~> 3.0` in order to allow for the BSON upgrade there.

adrianmacneil:
Awesome, thanks! :clap:

PR #34, opened 2015-06-04
Allow BSON::Regexp to delegate methods to Regexp

tubbo:
Ruby's Regexp method allows for a ton of methods that BSON::Regexp::Raw
does not define. This change lets any method not defined by the
BSON::Regexp::Raw class to be delegated to the underlying Regexp object
provided by BSON::Regexp::Raw#compile. It should cut down on boilerplate
code on the app level that allows you to use a BSON::Regexp::Raw as a Regexp,
by calling methods like match(), scan(), etc.

### DIFF
--- a/lib/bson/regexp.rb
+++ b/lib/bson/regexp.rb
@@ -120,6 +120,21 @@ module BSON
         @pattern = pattern
         @options = options
       end
+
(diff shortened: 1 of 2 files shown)


durran:
Oh nice. Thanks!

PR #35, opened 2015-06-14
add support for ActiveSupport Hash#slice method

john-denisov:
A common practice in rails is to `#slice` hash by array of symbols. `#slice` [expects](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/hash/slice.rb#L22) any conversion on key to happen in `#convert_key` method. That was made to support `HashWithIndifferentAccess`. [Implementation of `#convert_key` in HashWithIndifferentAccess](https://github.com/rails/rails/blob/3e36db4406beea32772b1db1e9a16cc1e8aea14c/activesupport/lib/active_support/hash_with_indifferent_access.rb#L258)

before this changes:
```ruby
bson_doc = BSON::Document.new(hello: 'world')
bson_doc.slice(:hello) # => {}
```
after this changes:
```ruby
bson_doc = BSON::Document.new(hello: 'world')
bson_doc.slice(:hello) # => {'hello' => 'world'}
```

### DIFF
--- a/lib/bson/document.rb
+++ b/lib/bson/document.rb
@@ -50,7 +50,7 @@ module BSON
     #
     # @since 2.0.0
     def [](key)
-      super(key.to_bson_normalized_key)
+      super(convert_key(key))
     end
 
     # Set a value on the document. Will normalize symbol keys into strings.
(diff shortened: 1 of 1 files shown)


durran:
Thanks!

PR #39, opened 2015-07-30
md5 hash the hostname in the native extension

hudon:
The native extension implementation for `next_object_id` doesn't match Ruby's. The latter hashes the hostname while the former does not. This patch fixes the native extension to also MD5 hash the hostname.

### DIFF
--- a/ext/bson/native.c
+++ b/ext/bson/native.c
@@ -49,12 +49,12 @@
 #endif
 
 /**
- * Define the max hostname length constant if nonexistant.
+ * Define the max hostname hash length constant if nonexistant.
  *
- * @since 2.0.0
+ * @since 3.2.0
  */
-#ifndef HOST_NAME_MAX
-#define HOST_NAME_MAX 256
+#ifndef HOST_NAME_HASH_MAX
+#define HOST_NAME_HASH_MAX 256
 #endif
 
 /**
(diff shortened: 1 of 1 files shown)


adrianmacneil:
For context (and seo for others running into this issue) this bug has been causing a large number of duplicate ID errors for us:

    The operation: #<Moped::Protocol::Command @length=81 @request_id=351 @response_to=0 @op_code=2004 @flags=[] @full_collection_name="db.$cmd" @skip=0 @limit=-1 @selector={:getlasterror=>1, :w=>1} @fields=nil> failed with error 11000: "E11000 duplicate key error index: db.collection.$_id_ dup key: { : ObjectId('55ba4acc69702d000100016a') }" See https://github.com/mongodb/mongo/blob/master/docs/errors.md for details about this error.

Investigating further, the hostname portion of the ObjectId from all servers was set to `ip-`:

    > BSON::ObjectId.from_string('55ba4acc69702d000100016a').to_bson[4..6]
    => "ip-"

This happened to be the prefix of all our machines' hostnames, which meant that if any two machines generated an ID in the same second with the same counter, we got an ObjectId clash.

durran:
Thanks. I'll get a release out quickly.

durran:
3.2.1 is out with this fix - thanks again.

PR #44, opened 2015-10-06
Fix Regexp.from_bson for data from SSL connections

Nielsomat:
This PR unifies parsing of regular expressions across SSL and non-SSL connections thus fixing [RUBY-1048](https://jira.mongodb.org/browse/RUBY-1048). From the issue description:

> When accessing mongodb through SSL, Regexp attributes can't be read from the database. https://github.com/mongodb/bson-ruby/blob/master/lib/bson/regexp.rb#L156 hangs indefinitely as the options are NULL_BYTE terminated.

> Note that this issue only occurs when connecting through SSL. On an unencrypted connection, options are indeed "0" terminated and the bug thus doesn't manifest itself using non-encrypted connections.

> The parsing of the options itself is also subject to differences between SSL and non-SSL connections.

> Test case in Rails:

> ```ruby
class TestRecord
  include Mongoid::Document
  field :regex, type: Regexp
end 
> 
> TestRecord.create(regex: //).reload # Will never return when connecting to MongoDB through SSL
```

### DIFF
--- a/lib/bson/regexp.rb
+++ b/lib/bson/regexp.rb
@@ -153,13 +153,13 @@ module BSON
       def from_bson(bson)
         pattern = bson.gets(NULL_BYTE).from_bson_string.chop!
         options = 0
-        while (option = bson.readbyte) != 0
+        while (option = bson.readbyte.chr) != NULL_BYTE
           case option
-          when 105 # 'i'
+          when "i"
             options |= ::Regexp::IGNORECASE
-          when 109, 115 # 'm', 's'
+          when "m", "s"
             options |= ::Regexp::MULTILINE
-          when 120 # 'x'
+          when "x"
             options |= ::Regexp::EXTENDED
           end
         end


estolfo:
Hi @Nielsomat 
Thanks so much for this. I've commented in the [jira ticket.](https://jira.mongodb.org/browse/RUBY-1048?focusedCommentId=1052584&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1052584) It'd be great if you could provide a test for the scenario your code addresses and rebase against master. Thanks!

estolfo:
Hi @Nielsomat 

Let me know if you're able to write a test for this and rebase against master so I can merge it. Thanks!

Emily

durran:
This is also fixed on my branch:

https://github.com/mongodb/bson-ruby/pull/43/files#diff-d0d0c547552be777373e58660aabaa99R153

Nielsomat:
@estolfo, this commit simply brings the #readbyte handling within Regexp in line with existing code such as https://github.com/mongodb/bson-ruby/blob/master/lib/bson/array.rb#L97 or https://github.com/mongodb/bson-ruby/blob/master/lib/bson/hash.rb#L77. As it is not explicitly tested there, I didn't include tests of my own. I have adapted the commit message accordingly.

It sounds like @durran's branch is the way to go in the long-term. For the short-term, I would suggest including this fix ASAP. The symptom of the bug isn't "just" invalid or incomplete data but rather the entire process hanging.

Right now, I am also working on a PR against mongodb/mongo-ruby-driver to address the underlying issue of the raw TCP and SSL socket behaving differently. I will reference this PR here once completed.

Nielsomat:
The root cause for this is now fixed in https://github.com/mongodb/mongo-ruby-driver/pull/700.

durran:
3.2.6 also has been released with this change. :)

Nielsomat:
:heart_eyes:

Excellent! Thanks, @durran!

PR #46, opened 2015-10-27
Support symbols for BSON::Document has_key?, has_value?

mtowers:
* Coerce passed symbols to strings for `has_key?` and `has_value?` methods
* Add standard aliases
* Update tests
* Increment version


durran:
Thanks - I have merged into master as this is an added feature.

mtowers:
Missed one case.  This same fix should be applied to `Document::delete()` as well.

PR #50, opened 2015-11-20
Serialize symbols to string types

durran:
We previously serialized symbols as strings, since they were deprecated, but still set the type byte in the serialized BSON to `0x0E`, which was the deprecated symbol type. This pull changes the serialized type to be `0x02`, which is string, while still maintaining the ability to *deserialize* symbols when the type in the BSON being decoded is `0x0E`.


estolfo:
looks good

PR #51, opened 2015-12-13
Initialize ObjectId counter to a random value

alethea:
The [ObjectId spec](https://docs.mongodb.org/v3.0/reference/object-id/) says the counter should be initialized to a random value. Initialize the counter using Ruby's `Kernel#rand` in both the native extension and the pure Ruby implementation.

### DIFF
--- a/ext/bson/native.c
+++ b/ext/bson/native.c
@@ -92,7 +92,7 @@ static char rb_bson_machine_id_hash[HOST_NAME_HASH_MAX];
 /**
  * The counter for incrementing object ids.
  */
-static unsigned int rb_bson_object_id_counter = 0;
+static unsigned int rb_bson_object_id_counter;
 
 /**
  * Initialize the native extension.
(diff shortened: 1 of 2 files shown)


estolfo:
Thanks @alethea 
Shouldn't we update this line as well? https://github.com/PillPack/bson-ruby/blob/random_object_id/ext/bson/native.c#L95

alethea:
I think it's cleaner to leave the initializer in `Init_native`. But I'm not a Ruby extension expert and I'm happy to move it.

estolfo:
You can leave the initialize line but don't have to initialize the value to 0. That seems less confusing as the value will never actually be 0.

alethea:
Done! The initialization line now declares the counter as `0x40c37f`.

estolfo:
How about if we leave that line as:
```c
static unsigned int rb_bson_object_id_counter;
```

alethea:
Done! Would you like me to squash my commits?

estolfo:
yes, please!

alethea:
Squashed. Thanks so much for the review.

estolfo:
Thanks so much @alethea !

alethea:
@estolfo Is there any chance we could get a v4.0.1 out with this in it?

PR #53, opened 2016-02-15
determine rubinius as well for generate object_id

edwardzhou:
Rubinius is real multi-threading as same as jruby. need to generate thread-safe object_id as well.

### DIFF
--- a/lib/bson/environment.rb
+++ b/lib/bson/environment.rb
@@ -33,7 +33,7 @@ module BSON
       defined?(JRUBY_VERSION)
     end
 
-    # Determin if we are using Rubinius or not.
+    # Determine if we are using Rubinius or not.
     #
     # @example Are we running with Rubinius?
     #   Environment.rubinius?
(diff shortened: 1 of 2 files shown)


estolfo:
Thanks for this @edwardzhou 
If you could just update the spelling of Determin to Determine in the documentation of the #rubinius? method, I can merge this. Thanks!

edwardzhou:
fixed. thanks

durran:
Thanks for the pull request, but the code path you have added will never get executed in Rubinius. Rubinius will be loading the C extensions, which will be generating the ObjectId via this native method: https://github.com/mongodb/bson-ruby/blob/master/ext/bson/native.c#L534

If there's an issue you have with the gem and Rubinius could you explain exactly what the problem is in a Jira issue in the Ruby project (https://jira.mongodb.org/browse/RUBY) and we can try to resolve it, but the code additions in this pull request would have no affect.

PR #59, opened 2016-04-25
RUBY-1092 Validate keys of hashes in an array

estolfo:
I noticed that we needed this set of changes when testing key validation with the driver.

The driver would try to serialize a write command like the following:
```ruby
{:insert=>"test", :documents=>[{"$testing"=>"value", :_id=>BSON::ObjectId('...')}], :ordered=>true, :writeConcern=>{:w=>1}}
```

The keys of the documents would not be validated because the top-level type is an Array. This pull request also applies key validation on hashes that are members of an Array so that we can actually use it from the driver.

### DIFF
--- a/lib/bson/array.rb
+++ b/lib/bson/array.rb
@@ -40,13 +40,13 @@ module BSON
     # @see http://bsonspec.org/#/specification
     #
     # @since 2.0.0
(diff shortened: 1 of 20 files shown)


durran:
:+1:

PR #60, opened 2016-05-17
Renamed extension from native to bson_native

genki:
Current extension name "native" is too loud in global namespace.
So I propose to add prefix "bson_" and change it to "bson_native".

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -35,7 +35,7 @@ if jruby?
 else
   require "rake/extensiontask"
   Rake::ExtensionTask.new do |ext|
-    ext.name = "native"
+    ext.name = "bson_native"
     ext.ext_dir = "ext/bson"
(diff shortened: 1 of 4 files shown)


estolfo:
Hi @genki 
What do you mean by "too loud"?
It's implied that it's the native bson extension, as the gem name is bson. Let me know why exactly you'd like a repetition of the gem name in the native file name.

genki:
I meant, we can `require "native"` anywhare. Isn't it annoying?

durran:
I have no problem changing the name - it doesn't break the public API. If we do, we should also change the java extensions as well, i.e. `NativeService` should probably become `BSONNativeService`.

genki:
@durran I think It's not necessary because the Java extension already has prefix `bson-`. I just tried to bring its manner to Ruby :)

PR #61, opened 2016-06-01
RUBY-1098 Decimal128 implementation II

estolfo:


lib/bson/decimal128.rb
```diff
@@ -0,0 +1,310 @@
...
+      # @example Get the message.
+      #   error.message
+      #
+      # @return [ String ] The error message.
+      #
+      # @since 4.1.0
+      def message
+        "Value out of range for Decimal128 representation. The exponent provided (#{@exponent}) " +
+          "must be at least #{Decimal128::MIN_EXPONENT} and no greater than " +
+          "#{Decimal128::MAX_EXPONENT}. The provided significand (#{@significand}) can only have " +
```
behackett: This isn't strictly true. With clamping, the max exponent is 6144, not 6111.

estolfo: ok, I'll update this message. thanks!

lib/bson/decimal128.rb
```diff
@@ -0,0 +1,310 @@
...
+# limitations under the License.
+
+require 'bigdecimal'
+require 'bson/decimal128/builder'
+
+module BSON
+
+  class Decimal128
+
+    # A Decimal128 is type 0x0D in the BSON spec.
```
bjori: No... Its 0x13

estolfo: thanks

jyemin:
Nit: Update all of your copyright messages to 2016.

ext/bson/bson_native.c
```diff
@@ -353,6 +372,27 @@ VALUE rb_bson_byte_buffer_put_cstring(VALUE self, VALUE string)
...
+ */
+VALUE rb_bson_byte_buffer_put_decimal128(VALUE self, VALUE low, VALUE high)
+{
+  byte_buffer_t *b;
+  const int64_t low64 = BSON_UINT64_TO_LE(NUM2ULL(low));
+  const int64_t high64 = BSON_UINT64_TO_LE(NUM2ULL(high));
+
+  TypedData_Get_Struct(self, byte_buffer_t, &rb_byte_buffer_data_type, b);
+  ENSURE_BSON_WRITE(b, 8);
+  memcpy(WRITE_PTR(b), (char*)&low64, 8);
```
behackett: The second parameter to memcpy is type void. No need to cast to char.

ext/bson/bson_native.c
```diff
@@ -353,6 +372,27 @@ VALUE rb_bson_byte_buffer_put_cstring(VALUE self, VALUE string)
...
+  b->write_position += 8;
+
+  ENSURE_BSON_WRITE(b, 8);
+  memcpy(WRITE_PTR(b), (char*)&high64, 8);
```
behackett: Same here.

lib/bson/decimal128.rb
```diff
@@ -0,0 +1,298 @@
...
+    # @return [ String ] The decimal128 as a string.
+    #
+    # @since 4.1.0
+    def to_s
+      @string ||= Builder::ToString.new(self).string
+    end
+    alias :to_str :to_s
+
+    # Get a BigDecimal object corresponding to this Decimal128.
+    # Note that the precision of the resulting BigDecimal is not guaranteed to exactly
```
behackett: Can you explain this? Ruby's BigDecimal is arbitrary precision.

johnnyshields: This comment is probably incorrect. I would expect that BigDecimal can exactly represent all possible Decimal128 objects, however, Decimal128 cannot represent all possible BigDecimal objects (BigDecimal is a superset of Decimal128)

estolfo: No, because specifying the precision is sometimes lost with BigDecimal objects. For example:

2.2.3 :004 > d = BigDecimal.new(2.00, 3)
 => #<BigDecimal:7fb43c0ed8d0,'0.2E1',9(27)>

johnnyshields: ah OK "the number of trailing zeroes". This is different than a (https://en.wikipedia.org/wiki/Loss_of_significance)["loss of significance"] I think a better wording would be:

> There is no loss of significance during conversion, however, the number of trailing zeroes of the resulting BigDecimal is not guaranteed to match that of the BSON Decimal128 object.

johnnyshields: It may be more accurate to say "truncating trailing zeroes is a special case of loss of significance"

estolfo: yes, I'll update the wording. Though you also have the issue of representation loss: for example, -NaN can be represented with a Decimal128 object but not with BigDecimal:

```ruby
> d = BSON::Decimal128.from_string("-NaN")
> d.to_big_decimal
=> #<BigDecimal:7fe49162a8b8,'NaN',9(9)>
```

johnnyshields: Ok so the -NaN case should be listed here as well. Again in 99% of practical use cases this is a very minor point, given that both -0 and -Infinity can be represented.

Anything else?

johnnyshields:
@estolfo I think it makes sense to coerce -NaN, sNaN etc. to NaN (or at least have a global config to do so, which is enabled by default). It's going to be a pain to deal with type conversion errors at runtime.

lib/bson/decimal128.rb
```diff
@@ -0,0 +1,310 @@
...
+        decimal = allocate
+        decimal.send(:set_bits, low, high)
+        decimal
+      end
+    end
+
+    # Raised when a Decimal128 is instantiated with a non-BigDecimal type.
+    #
+    # @since 4.1.0
+    class InvalidBigDecimal < RuntimeError; end
```
johnnyshields: Shouldn't this extend ArgumentError (same with InvalidString and InvalidRange)

lib/bson/decimal128.rb
```diff
@@ -0,0 +1,310 @@
...
+    # @param [ BigDecimal ] big_decimal The BigDecimal to use for
+    #   instantiating a Decimal128.
+    #
+    # @raise [ InvalidBigDecimal ] Raise error unless object argument is a BigDecimal.
+    #
+    # @since 4.1.0
+    def initialize(big_decimal)
+      raise InvalidBigDecimal.new unless big_decimal.is_a?(BigDecimal)
+      set_bits(*Builder::FromBigDecimal.new(big_decimal).bits)
+    end
```
johnnyshields: I would propose the following constructor which mirrors BigDecimal as closely as possible.

```
    def initialize(initial, digits = nil)
      big_decimal = BigDecimal.new(initial, digits)
      set_bits(*Builder::FromBigDecimal.new(big_decimal).bits)
    end
```

`initial` can be Integer, a Float, a Rational, a BigDecimal, or a String. Refer to BigDecimal constructor.

lib/bson/decimal128.rb
```diff
@@ -0,0 +1,307 @@
...
+    # @param [ BigDecimal ] big_decimal The BigDecimal to use for
+    #   instantiating a Decimal128.
+    #
+    # @raise [ InvalidBigDecimal ] Raise error unless object argument is a BigDecimal.
+    #
+    # @since 4.2.0
+    def initialize(big_decimal)
+      raise InvalidBigDecimal.new unless big_decimal.is_a?(BigDecimal)
+      set_bits(*Builder::FromBigDecimal.new(big_decimal).bits)
+    end
```
johnnyshields: I would propose the following constructor which mirrors BigDecimal as closely as possible.

```ruby
    def initialize(initial, digits = nil)
      big_decimal = BigDecimal.new(initial, digits)
      set_bits(*Builder::FromBigDecimal.new(big_decimal).bits)
    end
```

`initial` can be Integer, a Float, a Rational, a BigDecimal, or a String. `digits` is required for Float and Rational types. Refer to BigDecimal constructor.

lib/bson/decimal128.rb
```diff
@@ -0,0 +1,307 @@
...
+        decimal = allocate
+        decimal.send(:set_bits, low, high)
+        decimal
+      end
+    end
+
+    # Raised when a Decimal128 is instantiated with a non-BigDecimal type.
+    #
+    # @since 4.2.0
+    class InvalidBigDecimal < RuntimeError; end
```
johnnyshields: Shouldn't this extend ArgumentError (same with InvalidString and InvalidRange)

estolfo:
Hi @johnnyshields 
[Our spec](https://github.com/mongodb/specifications/blob/master/source/bson-decimal128/decimal128.rst) says that "Drivers MAY choose to support signed NaN (sNaN), along with sNaN with
diagnostic information." so I'm going to leave it that sNaN, -sNaN, and -NaN are coerced to a BigDecimal NaN. This is the current implementation.

johnnyshields:
@estolfo got it re: sNaN, great!

lib/bson/decimal128.rb
estolfo: that's a good idea, thanks!

johnnyshields:
@estolfo big picture question here.

In Mongoid, if I do:

```
class Car
  include Mongoid::Document

  field :wheels, type: BSON::Decimal128
end

car1 = Car.new
car2 = Car.new
car1.wheels = 2
car2.wheels = 3
car1.wheels + car2.wheels == 5
```

Will this work?

I'm a bit concerned here because other classes like BSON::Float are included into Float class itself. Decimal128 is not included into BigDecimal, and doesn't seem to implement arithmetic methods to allow it to be manipulated naturally within Ruby (unless I'm missing something).

estolfo:
You wouldn't be able to do that; no math operators are implemented for the BSON::Decimal128 object. See [this section](https://github.com/mongodb/specifications/blob/master/source/bson-decimal128/decimal128.rst#operator-overloading-and-math-on-decimal128-value-objects) of the spec.

lib/bson/decimal128.rb
estolfo: Perhaps...but I can't change InvalidString or InvalidRange to extend ArgumentError until a major version of the bson gem. I'm getting rid of this error anyway and letting BigDecimal raise if it cannot be instantiated with the args to #initialize.

johnnyshields:
@estolfo in that case, may we do:

```
  field :wheels, type: BigDecimal
```

And have it convert to/from BSON::Decimal128 automatically under the hood, with the known limitations where the two types don't align exactly?

Real-world users need a way to easily take advantage of the Decimal128 functionality without having to call conversion methods in 1000 places in their code, this is the core feature that makes Mongoid so attractive.

estolfo:
We would have to have a global config in that case, otherwise it wouldn't be backwards compatible for users who have been using BigDecimal already in Mongoid (as you known, BigDecimal objects are "serialized" and saved as strings in the db). I believe .NET has chosen to expose a global config like this in its ODM.

lib/bson/decimal128.rb
```diff
@@ -0,0 +1,311 @@
...
+    # @param [ Integer, Float, Rational, BigDecimal, String ] initial The initial value to use for
+    #   instantiating a Decimal128.
+    # @param [ Fixnum ] digits The number of significant digits, as a Fixnum.
+    #   If omitted or 0, the number of significant digits is determined from the initial value.
+    #
+    # @since 4.2.0
+    def initialize(initial, digits = nil)
+      big_decimal = BigDecimal.new(*[initial, digits].compact)
+      set_bits(*Builder::FromBigDecimal.new(big_decimal).bits)
+    end
```
johnnyshields: Just thinking a bit further here, you might want to do:

```ruby
def initialize(initial, digits = nil)
  if initial.is_a?(String)
    set_bits(*Builder::FromString.new(initial).bits)
  else
    big_decimal = BigDecimal.new(*[initial, digits].compact)
    set_bits(*Builder::FromBigDecimal.new(big_decimal).bits)
  end
end
```

In this manner you could handle the `"-sNaN"` etc cases.

johnnyshields:
@estolfo I agree with the idea of a global config, I do think that 99% of users (assuming they are on MongoDB 3.4) would want to enable the config.

durran:
I generally pull the c extension code from libbson, so maybe as a second pass for optimizations (especially on string operations) you might want to pull some code from there:

https://github.com/mongodb/libbson/blob/master/src/bson/bson-decimal128.c

lib/bson/decimal128.rb
```diff
@@ -0,0 +1,315 @@
...
+    #   instantiating a Decimal128.
+    # @param [ Fixnum ] digits The number of significant digits, as a Fixnum.
+    #   If omitted or 0, the number of significant digits is determined from the initial value.
+    #
+    # @since 4.2.0
+    def initialize(initial, digits = nil)
+      if initial.is_a?(String)
+        set_bits(*Builder::FromString.new(initial).bits)
+      else
+        big_decimal = BigDecimal.new(*[initial, digits].compact)
```
durran: Is it a significant overhead to first instantiate a `BigDecimal` vs having the builder handle all possibilities itself? If not, then I'd say this is fine... If it is, then I'd look into other options.

estolfo: Good point. I'll do some benchmarking.

estolfo: @durran I've updated the #initialize method to create a Decimal128 directly from an Integer and Float, instead of instantiating a BigDecimal first. Benchmarking showed that this was significantly faster.

lib/bson/decimal128.rb
```diff
@@ -107,11 +107,18 @@ def ==(other)
...
-      else
-        big_decimal = BigDecimal.new(*[initial, digits].compact)
-        set_bits(*Builder::FromBigDecimal.new(big_decimal).bits)
+      case initial
+        when String
+          set_bits(*Builder::FromString.new(initial).bits)
+        when Integer
+          set_bits(*Builder::FromInteger.new(initial).bits)
+        when Float
+          set_bits(*Builder::FromFloat.new(initial).bits)
```
johnnyshields: `digits` param should be required for Float

estolfo: but I extract the number of sig digits here, is it really necessary? https://github.com/mongodb/bson-ruby/pull/61/files/2ef994a4d637fb1187d598db63a5277139b0e5ef..b3a4cdd1a82c43eb431558084647b9a27e535453#diff-87a1887ce145bf861a2df29a0ac56b82R286

johnnyshields: Float can have all sorts of precision related errors, for example try this in IRB:

139.25 + 74.79

We should at least make `digits` optional if not required. (BigDecimal requires it)

estolfo: I've added support for specifying digits of precision when passing a Float to Decimal128#new  https://github.com/mongodb/bson-ruby/pull/61/commits/37bd9dde968073e7833a6d529a3ec4fda3a01272

estolfo:
After speaking with other members of the drivers team, I've decided to only accept a BigDecimal object to the #initialize method of BSON::Decimal128. For other objects like Float, I would have to call #to_s on them to get the digits of precision and do a lot of parsing. The best way to ensure that the BSON::Decimal128 has the desired digits of precision is to instantiate the object using Decimal128.from_string("...")
Please see the spec [here](https://github.com/mongodb/specifications/blob/master/source/bson-decimal128/decimal128.rst#from-string-representation) for how we create a Decimal128 from a String.

johnnyshields:
@estolfo firstly thanks for the great work.

I disagree with the decision to only accept BigDecimal in the constructor, but can live it with as long as Mongoid provides a way that `type: BigDecimal` gets seamlessly/opaquely mapped to `Decimal128` in MongoDB. (In fact I hope to never directly instantiate a `BSON::Decimal128` in my app, ever!)

estolfo:
Hi @johnnyshields 

We can always add more functionality to the constructor later if a lot of people ask for it though you still have the method Decimal128.from_string(...) to turn a String into a Decimal128 object.
For Mongoid, what we can do is change the #demongoize and #mongoize methods to handle String or BigDecimal/Decimal128 input/output depending on a global config.

estolfo:
@johnnyshields See this pull request: https://github.com/mongodb/bson-ruby/pull/66
I realized it would be simpler to handle Strings in the Decimal128 constructor instead of having to called "#from_string explicitly when "demongoizing" Mongoid BigDecimals previously serialized as Strings in the db.

PR #62, opened 2016-07-07
Support for serializing OpenStruct objects

adimitrov:
Because OpenStruct behaves like a hash, when used, bson internally calls
`.bson_type` on it, which is undefined and always returns `nil`
and causes a segfault.

This PR makes it so OpenStruct objects get converted a Hash and then serialized 

Example Testcase:

```
class Foo
   include Mongoid::Document
end

> Foo.find_by(bar: OpenStruct.new)

/Users/alex/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bson-4.1.1/lib/bson/hash.rb:45: [BUG] Segmentation fault at 0x00000000000009
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin15]
```


estolfo:
Hi @adimitrov 
This should go into Mongoid instead, and you can add a #mongoize method to achieve the same functionality. Would you mind making the pull request to Mongoid instead? Thanks

estolfo:
On second thought, if there is a segmentation fault caused by this, it may be necessary to add to the bson gem. I'll look into it and get back to you. Thanks!

lib/bson/open_struct.rb
```diff
@@ -0,0 +1,47 @@
...
+    #   OpenStruct.new({ "field" => "value" }).to_bson
+    #
+    # @return [ String ] The encoded string.
+    #
+    # @see http://bsonspec.org/#/specification
+    #
+    # @since 2.0.0
+    def to_bson(buffer = ByteBuffer.new, validating_keys = Config.validating_keys?)
+      if RUBY_VERSION < "2.0.0"
+        marshal_dump.dup
```
estolfo: why are you calling #dup here?

lib/bson/open_struct.rb
```diff
@@ -0,0 +1,47 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module BSON
+  # Injects behaviour for encoding OpenStruct objects using hashes
+  # to # raw bytes as specified by the BSON spec.
+  #
+  # @see http://bsonspec.org/#/specification
+  #
+  # @since 2.0.0
```
estolfo: this needs to be updated to 4.2.0

lib/bson/open_struct.rb
```diff
@@ -0,0 +1,47 @@
...
+    # Get the OpenStruct as encoded BSON.
+    #
+    # @example Get the hash as encoded BSON.
+    #   OpenStruct.new({ "field" => "value" }).to_bson
+    #
+    # @return [ String ] The encoded string.
+    #
+    # @see http://bsonspec.org/#/specification
+    #
+    # @since 2.0.0
```
estolfo: same here - update to 4.2.0

lib/bson/open_struct.rb
```diff
@@ -0,0 +1,47 @@
...
+    def to_bson(buffer = ByteBuffer.new, validating_keys = Config.validating_keys?)
+      if RUBY_VERSION < "2.0.0"
+        marshal_dump.dup
+      else
+        to_h
+      end.to_bson(buffer, validating_keys)
+    end
+
+    def bson_type
```
estolfo: this needs documentation

lib/bson/open_struct.rb
```diff
@@ -0,0 +1,47 @@
...
+    # @example Get the hash as encoded BSON.
+    #   OpenStruct.new({ "field" => "value" }).to_bson
+    #
+    # @return [ String ] The encoded string.
+    #
+    # @see http://bsonspec.org/#/specification
+    #
+    # @since 2.0.0
+    def to_bson(buffer = ByteBuffer.new, validating_keys = Config.validating_keys?)
+      if RUBY_VERSION < "2.0.0"
```
durran: Have we discussed dropping 1.9.x support? EOL for 1.9.3 was a year and a half ago. (https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/) If we can drop 1.9.3 support we can remove the marshal dump here...

estolfo: We can definitely look into that but it would have to go in a major release.

estolfo: also we need to check for JRuby here. JRuby < 9.0 always returns nil from OpenStruct#to_h

```ruby
jruby-1.7.22 :003 > o = OpenStruct.new(:country => "Australia")
 => #<OpenStruct country="Australia"> 
jruby-1.7.22 :004 > o.to_h
 => nil 
```

JRuby 9.0:

```ruby
jruby-9.1.2.0 :003 > o = OpenStruct.new(:country => "Australia")
 => #<OpenStruct country="Australia"> 
jruby-9.1.2.0 :004 > o.to_h
 => {:country=>"Australia"} 
```

estolfo: Nevermind. ``RUBY_VERSION`` works in jRuby as well for this.

PR #64, opened 2016-07-29
Avoid ruby warnings

qbantek:
Removes this from the logs => warning: instance variable @validating_keys not initialized


estolfo:
thanks!

PR #65, opened 2016-08-11
[for 4.2.0] Add BSON::ObjectId#to_time as alias to #generation_time

oss92:
Goes well with `#from_time` and `#to_s`


estolfo:
Thanks @oss92 
This will be merged in for the next minor version, as it's a new method. I've opened a ticket to track adding it in for bson 4.2.0: https://jira.mongodb.org/browse/RUBY-1138

PR #66, opened 2016-09-15
RUBY-1098 Support a String passed to Decimal128#new

estolfo:
When looking into how Mongoid would handle converting BigDecimal objects previously "serialized" as Strings in the db to Decimal128 objects, I realized it would be simpler to handle Strings in the Decimal128 constructor. Otherwise, the method Decimal128#from_string would have to be used explicitly.

### DIFF
--- a/lib/bson/decimal128.rb
+++ b/lib/bson/decimal128.rb
@@ -239,6 +239,8 @@ module BSON
 
     # Raised when trying to create a Decimal128 from an object that is neither a String nor a BigDecimal.
     #
+    # @api private
+    #
     # @since 4.2.0
     class InvalidArgument < ArgumentError
 
@@ -263,6 +265,8 @@ module BSON
     # Raised when trying to create a Decimal128 from a string with
     #   an invalid format.
     #
+    # @api private
+    #
     # @since 4.2.0
     class InvalidString < RuntimeError
 
@@ -285,6 +289,8 @@ module BSON
     end
 
     # Raised when the exponent or significand provided is outside the valid range.
+    #
+    # @api private
     #
     # @since 4.2.0
     class InvalidRange < RuntimeError
--- a/lib/bson/decimal128.rb
+++ b/lib/bson/decimal128.rb
@@ -54,11 +54,6 @@ module BSON
     # @since 4.2.0
     NATIVE_TYPE = BigDecimal.freeze
 
-    # The error message if neither a String nor a BigDecimal are passed to #new.
-    #
-    # @since 4.2.0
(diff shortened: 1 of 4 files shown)


johnnyshields:
Hi Emily, this is good, however if constructor supports string and BigDecimal it begs the question why it doesn't support Integer and Float (like we had on the previous PR until it got reverted.) I think the broadest possible type coverage will make it easiest to work with.

estolfo:
The biggest reason is, for the case of Float, there are no auxiliary methods like there are for BigDecimal to get sign, significant digits, unscaled value, exponent, etc. I'm therefore forced to have to call #to_s on the object and parse the String. I don't know what rules Float follows to print to a String so I'm afraid the String representation I would parse would not exactly represent the Float passed in. I'd rather the user pass in the String him/herself.

lib/bson/decimal128.rb
```diff
@@ -54,6 +54,11 @@ class Decimal128
     # @since 4.2.0
     NATIVE_TYPE = BigDecimal.freeze
 
+    # The error message if neither a String nor a BigDecimal are passed to #new.
+    #
+    # @since 4.2.0
```
estolfo: Testing review

estolfo: test

jyemin: Is this part of the public API?  Does it need to be?

spec/bson/decimal128_spec.rb
```diff
@@ -26,6 +26,15 @@
 
   describe '#initialize' do
 
+    context 'when the argument is neither a BigDecimal or String' do
```
jyemin: I like the test.

lib/bson/decimal128.rb
estolfo: Yes, this is the public API and you can't make a private constant in Ruby. Thanks for pointing this out though, it's kind of an awkward design and inconsistent with the rest of the bson gem. I'll figure out how to do this differently.

estolfo: Resolved in latest commit.

jyemin: lgtm

johnnyshields:
@estolfo so, firstly I think we should separate the cases of Integer and Float. Integer is trivially easy and I think it should be supported.

For Float, I can think of three approaches:
1. Do a bit-wise conversion (do not convert to String) directly to Decimal128
2. Convert to String internally first, then to Decimal128
3. Convert to BigDecimal internally first, then to Decimal128

Would you accept a PR for any of the above approaches?

PR #69, opened 2016-10-05
RUBY-1144 Implement get_document & get_array in C

fcheung:
This  also hardcodes mapping of ints & doubles as an example of the speed improvement this brings.

MAPPINGS has changed from string keys to integer keys so that we don't
have to allocate a new string on every single field in order to do the lookup

See https://jira.mongodb.org/browse/RUBY-1144


ext/bson/bson_native.c
```diff
@@ -180,6 +225,173 @@ VALUE rb_bson_byte_buffer_initialize(int argc, VALUE *argv, VALUE self)
...
+
+  position = READ_SIZE(b);
+  bson_byte_buffer_put_int32(b, 0);
+
+
+  array_element = RARRAY_PTR(array);
+
+  for(int32_t index=0; index < RARRAY_LEN(array); index++, array_element++){
+    char key_string[16];
+    snprintf(key_string, sizeof(key_string), "%d", index);
```
durran: https://github.com/mongodb/libbson/blob/master/src/bson/bson-keys.c#L24

fcheung: Thanks for the tip.

ext/bson/bson_native.c
```diff
@@ -87,6 +97,27 @@ static const rb_data_type_t rb_byte_buffer_data_type = {
   { NULL, rb_bson_byte_buffer_free, rb_bson_byte_buffer_memsize }
 };
 
+static VALUE bson_byte_buffer_get_int32(byte_buffer_t *b);
```
fcheung: sounds good

ext/bson/bson_native.c
```diff
@@ -61,6 +69,8 @@ static VALUE rb_bson_byte_buffer_get_double(VALUE self);
 static VALUE rb_bson_byte_buffer_get_int32(VALUE self);
 static VALUE rb_bson_byte_buffer_get_int64(VALUE self);
 static VALUE rb_bson_byte_buffer_get_string(VALUE self);
+static VALUE rb_bson_byte_buffer_get_document(VALUE self);
```
fcheung: ok

ext/bson/bson_native.c
```diff
@@ -180,6 +225,173 @@ VALUE rb_bson_byte_buffer_initialize(int argc, VALUE *argv, VALUE self)
   return self;
 }
 
+static int is_int32(int64_t i64){
```
fcheung: ok

fcheung:
Hi,

thanks for the feedback - very helpful. I get what you're saying about having lots of the put/get functions split in 2. As well as TypedData_Get_Struct calls, I was also trying to avoid cases where I'd have to create a ruby object to wrap an int, string etc that then gets passed to a function that unwraps the ruby object to get the previous value back.

I tinkered with undoing some of that and it does seem like there is a performance benefit to it - perhaps 10% or so. It would be cleaner if it could be avoided (and then functions that need to be able to fall back to pure ruby implementations (like put_field) wouldn't need to carry arround both the ruby object and the byte_buffer_t. I can't think of a way to structure it differently at the moment that would be easier to work with though.

ext/bson/bson_native.c
```diff
@@ -69,11 +79,14 @@ static VALUE rb_bson_byte_buffer_put_double(VALUE self, VALUE f);
...
 static VALUE rb_bson_byte_buffer_put_int64(VALUE self, VALUE i);
 static VALUE rb_bson_byte_buffer_put_string(VALUE self, VALUE string);
+static VALUE rb_bson_byte_buffer_put_hash(VALUE self, VALUE hash, VALUE validating_keys);
+static VALUE rb_bson_byte_buffer_put_array(VALUE self, VALUE array, VALUE validating_keys);
 static VALUE rb_bson_byte_buffer_read_position(VALUE self);
 static VALUE rb_bson_byte_buffer_replace_int32(VALUE self, VALUE index, VALUE i);
 static VALUE rb_bson_byte_buffer_rewind(VALUE self);
 static VALUE rb_bson_byte_buffer_write_position(VALUE self);
 static VALUE rb_bson_byte_buffer_to_s(VALUE self);
+
```
xdg: extra line

ext/bson/bson_native.c
```diff
@@ -180,6 +227,283 @@ VALUE rb_bson_byte_buffer_initialize(int argc, VALUE *argv, VALUE self)
...
+      pvt_put_byte(b, 1);
+      break;
+    case T_FALSE:
+      pvt_put_byte(b, 0);
+      break;
+    case T_HASH:
+      rb_bson_byte_buffer_put_hash(rb_buffer, val, validating_keys);
+      break;
+    default:{
+      rb_funcall(val, rb_intern("to_bson"), 2, rb_buffer,validating_keys);
```
xdg: nit: missing space after last comma

ext/bson/bson_native.c
```diff
@@ -180,6 +227,283 @@ VALUE rb_bson_byte_buffer_initialize(int argc, VALUE *argv, VALUE self)
...
+      pvt_put_bson_key(b, key, validating_keys);
+      break;
+    case T_SYMBOL:
+      pvt_put_bson_key(b, rb_sym_to_s(key), validating_keys);
+      break;
+    default:
+      rb_bson_byte_buffer_put_cstring(buffer, rb_funcall(key, rb_intern("to_bson_key"),1,validating_keys));
+  }
+
+  pvt_put_field(b, buffer, val,validating_keys);
```
xdg: nit: missing comma (and I note that 324 is inconsistent, too)

ext/bson/bson_native.c
```diff
@@ -180,6 +227,283 @@ VALUE rb_bson_byte_buffer_initialize(int argc, VALUE *argv, VALUE self)
...
+  size_t position = 0;
+  size_t new_position = 0;
+  int32_t new_length = 0;
+
+  TypedData_Get_Struct(self, byte_buffer_t, &rb_byte_buffer_data_type, b);
+  Check_Type(hash, T_HASH);
+
+  position = READ_SIZE(b);
+
+  pvt_put_int32(b, 0);
```
xdg: On reflection, I'd like a comment here like `/* insert length placeholder */`

ext/bson/bson_native.c
```diff
@@ -180,6 +227,283 @@ VALUE rb_bson_byte_buffer_initialize(int argc, VALUE *argv, VALUE self)
...
+  byte_buffer_t *b = NULL;
+  size_t new_position = 0;
+  int32_t new_length = 0;
+  size_t position = 0;
+  VALUE *array_element = NULL;
+  TypedData_Get_Struct(self, byte_buffer_t, &rb_byte_buffer_data_type, b);
+  Check_Type(array, T_ARRAY);
+
+  position = READ_SIZE(b);
+  pvt_put_int32(b, 0);
```
xdg: Ditto: please add a comment like `/* insert length placeholder */`

ext/bson/bson_native.c
```diff
@@ -235,10 +567,22 @@ VALUE rb_bson_byte_buffer_get_cstring(VALUE self)
   ENSURE_BSON_READ(b, length);
   string = rb_enc_str_new(READ_PTR(b), length, rb_utf8_encoding());
   b->read_position += length + 1;
+
```
xdg: extra line

ext/bson/bson_native.c
```diff
@@ -259,13 +603,20 @@ VALUE rb_bson_byte_buffer_get_decimal128_bytes(VALUE self)
...
+VALUE pvt_get_double(byte_buffer_t *b)
+{
+  double d;
+
   ENSURE_BSON_READ(b, 8);
   memcpy(&d, READ_PTR(b), 8);
   b->read_position += 8;
   return DBL2NUM(BSON_DOUBLE_FROM_LE(d));
+
```
xdg: extra line

ext/bson/bson_native.c
```diff
@@ -319,6 +687,71 @@ VALUE rb_bson_byte_buffer_get_string(VALUE self)
...
 
+
+VALUE rb_bson_byte_buffer_get_hash(VALUE self){
+  VALUE doc = Qnil;
+  byte_buffer_t *b;
+  char type;
+  VALUE cDocument = rb_const_get(rb_const_get(rb_cObject, rb_intern("BSON")), rb_intern("Document"));
+  TypedData_Get_Struct(self, byte_buffer_t, &rb_byte_buffer_data_type, b);
+
+  /* skip length */
```
xdg: I'd prefer to see this validate the length -- at confirming (a) the length is less than the remaining length of the buffer and (b) there's a null byte at the end.

ext/bson/bson_native.c
```diff
@@ -319,6 +687,71 @@ VALUE rb_bson_byte_buffer_get_string(VALUE self)
...
+}
+
+VALUE rb_bson_byte_buffer_get_array(VALUE self){
+  byte_buffer_t *b;
+  VALUE array = Qnil;
+  char type;
+  TypedData_Get_Struct(self, byte_buffer_t, &rb_byte_buffer_data_type, b);
+
+  ENSURE_BSON_READ(b, 4);
+  array = rb_ary_new2(pvt_get_int32(b));
```
xdg: The "length" of a BSON array is a document length, just like the document/hash case.  It's not the number of elements.

xdg: Also, shouldn't this need to increment the read position, like line 700?

lib/bson/hash.rb
```diff
@@ -73,13 +77,17 @@ module ClassMethods
...
-        hash = Document.allocate
-        buffer.get_int32 # Throw away the size.
-        while (type = buffer.get_byte) != NULL_BYTE
-          field = buffer.get_cstring
-          hash.store(field, BSON::Registry.get(type, field).from_bson(buffer))
+        if buffer.respond_to?(:get_hash)
+          buffer.get_hash
+        else
+          hash = Document.allocate
+          buffer.get_int32 # Throw away the size.
```
xdg: Ditto about size validation.

lib/bson/array.rb
```diff
@@ -93,13 +97,17 @@ module ClassMethods
...
-        array = new
-        buffer.get_int32 # throw away the length
-        while (type = buffer.get_byte) != NULL_BYTE
-          buffer.get_cstring
-          array << BSON::Registry.get(type).from_bson(buffer)
+        if buffer.respond_to?(:get_array)
+          buffer.get_array
+        else
+          array = new
+          buffer.get_int32 # throw away the length
```
xdg: I realize the original throws away size, but as I mentioned, I'd prefer to see validation.

ext/bson/bson_native.c
fcheung: Thanks - that was causing excessive memory consumption. *_get_int32 increment the read position for you

fcheung:
I've cleanup up the whitespace & added the validation you suggest. The ruby version of that validation is incomplete: it doesn't check for the null terminating byte because the api exposed to ruby doesn't have a method for fetching a byte at arbitrary offset. This code branch is only followed if you're on jruby and I feel that modifying the java implementation of ByteBuffer is probably beyond the scope of this PR (not to mention my skill).

fcheung:
Hmm, the length method on jruby appears to behave differently so my attempts to validate it failed

fcheung:
Specifically, this is the behaviour I see:

MRI:
```
BSON::ByteBuffer.new({"a": 1}.to_bson.to_s).length #=> 12
```
I get the length of the buffer, so I can validate that based on the current position and the array/hash length I have just read

jruby-9.1.8.0:
```
BSON::ByteBuffer.new({"a": 1}.to_bson.to_s).length #=> 0
```
So there's not much I can do. It seems like most of the length validation stuff in the java extension are no-ops and instead relies on the guarantees the underlying java.nio.ByteBuffer provides in order to prevent out of bounds access.

ext/bson/bson_native.c
```diff
@@ -110,6 +109,7 @@ static VALUE pvt_get_boolean(byte_buffer_t *b);
 static VALUE pvt_read_field(byte_buffer_t *b, VALUE rb_buffer, uint8_t type);
 static void pvt_replace_int32(byte_buffer_t *b, int32_t position, int32_t newval);
 static void pvt_skip_cstring(byte_buffer_t *b);
+static void pvt_validate_length(byte_buffer_t *b);
```
xdg: Nice refactoring.  Thank you!

estolfo:
Thank you so much @xdg and @fcheung for all your work and review on this.

As a final step, I'm going to run our driver and mongoid tests against the gem built with these changes and let you know what the outcome is.

estolfo:
I built the gem and ran our driver test suite using it locally and got a few errors. @fcheung would you mind doing the same and seeing if you know what the issue is? (``gem build bson.gemspec``, then install the .gem output and run the driver test suite)

Most of them are this error:

```
      expected Mongo::Error::BulkWriteError, got #<RangeError: Buffer should have contained null terminator at 175 but contained > with backtrace:
         # /Users/emily/.rvm/gems/ruby-2.4.1/gems/bson-4.2.2/lib/bson/hash.rb:81:in `get_hash'
         # /Users/emily/.rvm/gems/ruby-2.4.1/gems/bson-4.2.2/lib/bson/hash.rb:81:in `from_bson'
         # ./lib/mongo/dbref.rb:104:in `from_bson'
         # ./lib/mongo/protocol/serializers.rb:175:in `deserialize'
         # ./lib/mongo/protocol/message.rb:283:in `block in deserialize_array'
         # ./lib/mongo/protocol/message.rb:283:in `times'
         # ./lib/mongo/protocol/message.rb:283:in `deserialize_array'
         # ./lib/mongo/protocol/message.rb:136:in `block in deserialize'
         # ./lib/mongo/protocol/message.rb:134:in `each'
         # ./lib/mongo/protocol/message.rb:134:in `deserialize'
         # ./lib/mongo/server/connectable.rb:102:in `block in read'
```

fcheung:
I believe I've fixed those errors. I'd normally have written a failing spec first, but I'm not entirely sure where. spec/bson mostly has single objects, whereas you need something a bit more complicated to exhibit this behaviour. There's the corpus tests, but as far as I can tell the point there is that those are shared across all implementations so there's no sense in me just adding specs there.

Any guidance?

estolfo:
Could you add a more complicated example it to the document_spec?

fcheung:
FYI, the last bug I fixed had been giving this branch an unfair advantage when deserializing documents. Still looks to be about 2x-2.5x faster on the decode benchmarks and 2.5x-3x faster on the encode ones.

estolfo:
Here is the latest on evergreen: https://evergreen.mongodb.com/version/bson_ruby_test_21aae6d0f0f96fabf071286175a3d1984502bcb8

PR #73, opened 2016-12-08
RUBY-1189 Deserialize an int64 to a Fixnum, not Bignum on jRuby

estolfo:


spec/bson/int64_spec.rb
```diff
@@ -60,6 +60,53 @@
...
+
+      let(:integer) { BSON::Integer::MAX_32BIT + 1 }
+
+      let(:bson) do
+        BSON::ByteBuffer.new(integer.to_bson.to_s)
+      end
+
+      context "when on JRuby", if: BSON::Environment.jruby? do
+
+        it "desserializes to a Fixnum object" do
```
estolfo: thanks

PR #74, opened 2016-12-13
RUBY-1173 Force unpacking Q< and update big endian byte order check

estolfo:


ext/bson/native-endian.h
```diff
@@ -26,6 +26,16 @@
 #define BSON_BIG_ENDIAN    4321
 #define BSON_LITTLE_ENDIAN 1234
 
+/* See a similar check in ruby's sha2.h */
+# ifndef BSON_BYTE_ORDER
```
behackett: It doesn't look like BSON_BYTE_ORDER will ever be defined here. It looks like this block was in the right place before.

estolfo: ok, thanks

PR #81, opened 2017-04-20
Improve Ruby Symbol handling

ukblewis:
By removing calls to `.to_s` but instead handling in C and Java extensions, we can expect a significant performance boost for users doing large inserts with hashes with symbol keys and pure symbol values.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,7 @@ group :development, :test do
   gem 'rspec', '~> 3.2'
   gem 'rake-compiler'
   gem 'ruby-prof', :platforms => :mri
-
+  
   if ENV['CI']
     gem 'mime-types', '1.25' # v2.0+ does not support ruby 1.8
   else
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -492,6 +492,9 @@ axes:
       - id: "rhel55-32"
         display_name: "RHEL 5.5 32-bit"
         run_on: rhel55-32
+      - id: "ubuntu1604-arm64-small"
+        display_name: "Ubuntu 16.04 ARM64"
+        run_on: ubuntu1604-arm64-small
   - id: "all-rubies"
     display_name: Ruby Version
     values:
--- a/.evergreen/.evg.yml
+++ b/.evergreen/.evg.yml
@@ -489,6 +489,9 @@ axes:
       - id: "rhelpower"
         display_name: "RHEL 7.1 PowerPC"
         run_on: rhel71-power8-test
+      - id: "rhel55-32"
+        display_name: "RHEL 5.5 32-bit"
+        run_on: rhel55-32
   - id: "all-rubies"
     display_name: Ruby Version
     values:
@@ -496,10 +499,10 @@ axes:
         display_name: ruby-1.9
         variables:
            RVM_RUBY: "ruby-1.9"
-      - id: "ruby-2.0"
-        display_name: ruby-2.0
+      - id: "ruby-2.2"
+        display_name: ruby-2.2
         variables:
-           RVM_RUBY: "ruby-2.0"
+           RVM_RUBY: "ruby-2.2"
       - id: "ruby-2.3"
         display_name: ruby-2.3
         variables:
@@ -527,10 +530,6 @@ axes:
         display_name: ruby-2.4.1
         variables:
            RVM_RUBY: "ruby-2.4.1"
-      - id: "jruby-9.1"
-        display_name: jruby-9.1
-        variables:
-           RVM_RUBY: "jruby-9.1"
 
 buildvariants:
 - matrix_name: "tests-all-rubies-all-os"
@@ -540,8 +539,6 @@ buildvariants:
      - name: "test"
 - matrix_name: "tests-special-os"
   matrix_spec: { latest-stable-rubies: "*", special-os: "*" }
-  exclude_spec: { latest-stable-rubies: "jruby-1.7", special-os: "*" }
-  exclude_spec: { latest-stable-rubies: "jruby-9.1", special-os: "*" }
   display_name: "${latest-stable-rubies}, ${special-os}"
   tasks:
      - name: "test"
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -15,9 +15,9 @@ source ~/.rvm/scripts/rvm
 export JAVACMD=/opt/java/jdk8/bin/java
 export PATH=$PATH:/opt/java/jdk8/bin
 
-#if [ "$RVM_RUBY" == "ruby-head" ]; then
+if [ "$RVM_RUBY" == "ruby-head" ]; then
   rvm reinstall $RVM_RUBY
-#fi
+fi
 
 rvm use $RVM_RUBY
 gem install bundler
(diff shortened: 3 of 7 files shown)


ukblewis:
@estolfo Can I please have some feedback on this PR? I am not sure how to approach the tests that fail on older Ruby versions.

estolfo:
Hi @ukblewis 
Thank you very much for your work on this!
I still need some time to review the pull request, but in the meantime, I'm running our tests on this patch [here.](https://evergreen.mongodb.com/version/58fe14552fbabe516c005771)
Also, it looks like the memory_profiler gem version you added in the Gemfile isn't supported on 1.9. You can see the errors in the [travis tests](https://travis-ci.org/mongodb/bson-ruby/jobs/223775334). Would you mind resolving that?
Thanks!
Emily

ukblewis:
Hi Emily,
  

  
Thank you for your reply! I'm excited to see the results of this! :)
  

  
I added that gem for testing the memory benefits from the commit - I looked into a way to not load the Gem in older Ruby versions and only run the performance test in newer versions, but didn't find anything - do you have any suggestions?
  

  

  
  
  
  
Warmest regards,
  

  
Ben
  
  
  
  
  

  
>

estolfo:
Hi @ukblewis 
I think you can check the ruby version in the Gemfile and install the gem only for the version of Ruby in which it's supported. See [this](https://github.com/mongodb/mongoid/blob/5.2.0-stable/Gemfile#L4-L6) for an example.
Also, how about putting the profiling into a rake task? You could add it to Rakefile.

We also have a failure on our testing platform, on jruby-9.1, RHEL 7.0. You can see it here: https://evergreen.mongodb.com/task/bson_ruby_tests_all_rubies_all_os__all_os~rhel70_all_rubies~jruby_9.1_test_496f2bab6bef36678cc93c8c05cc7696ab1caef9_17_04_24_15_06_26
Let me know if you have any ideas.

Thanks
Emily

ukblewis:
Hi @estolfo 

So I've updated the Gemfile and my script to be more usable. I haven't added a Rake task since I don't anticipate it being something used in the long term - in fact - I'm happy to migrate the script to a Gist instead - it was simply to show the performance gains of this whole PR.

I'm waiting for the tests to run in Travis - but I expect them to pass. I can't access your Evergreen since I don't work for MongoDB (I work for Twist Bioscience - we've spoken a few times before).

I also have another thing that I'd like to point out that I don't know where to put. I have noticed that the MongoDB Ruby driver Gem merges IDs into document hashes if they aren't there. Since this uses the non-mutating `merge` method and not the mutating `merge!` method, it creates a copy of your document. If you are inserting a lot of documents - this can result in allocating a huge amount of memory and can take a really long time to execute. I think that either you should document that the driver encourages users to generate their own IDs for better performance or give users a flag to state that they require their documents unchanged and default to `merge!` instead of `merge`.


Thank you for your help! I'm looking forward to your feedback :)

Ben

estolfo:
Hi @ukblewis 

(Of course I remember speaking with you before = )

For the performance profiling I think it would be best to put it into a gist that we can use to justify these changes. I agree that it's not something we would need to have in the gem, longterm.

Sorry, I forgot our patch runs on Evergreen aren't publicly accessible. I'm running the tests again and I'll send you a gist of the output if it fails.

As for the ids, in our CRUD spec, it says that we add the identifier if there isn't one in the document. https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst
I think the best thing to do is make a note about adding one's own identifier to documents for performance and memory benefits. Our documentation is here: https://github.com/mongodb/mongo-ruby-driver/tree/master/docs if you like to open a pull request to add the note. Otherwise, I can add it. Let me know what you prefer!

Thanks again
Emily

ukblewis:
Hi @estolfo ,

Great :)

So I've removed the symbol performance testing file from the repository and put it into this GIST: https://gist.github.com/ukblewis/c5b1b89967db6d6f5f1a93848f052e03

With regards to the IDs, I'll be happy to update the documentation - I'll take a look now :)


Thank you!

Ben

Gemfile
```diff
@@ -12,6 +12,10 @@ group :development, :test do
   gem 'rspec', '~> 3.2'
   gem 'rake-compiler'
   gem 'ruby-prof', :platforms => :mri
+  gem 'benchmark-ips'
+  if RUBY_VERSION >= "2.1"
+    gem 'memory_profiler', :platforms => :mri
+  end
```
estolfo: We probably don't need this now that the benchmarking is in a gist, right?

ukblewis: We don't need it for the content of the Gem - yeah that is correct - should I add this in a comment or something to the gist and remove it from here?

estolfo: You can remove it since we don't need it for gem development. Thanks!

ukblewis:
Is this eligible for merge now?

ukblewis:
@estolfo I believe this is ready for merge. I haven't yet got to the documentation of the MongoDB Ruby driver as discussed - hopefully I'll get to that in the next few days.

estolfo:
ok, thanks @ukblewis

estolfo:
Ok, I've run our tests and this is passing. I was just wondering if you could squash it into a single commit?

ukblewis:
@estolfo I've just squashed it... not sure how it ended up as two commits :/ hope that I didn't do anything wrong - I'm not too comfortable with squashing Git commits.

estolfo:
No worries, I think you need to rebase against this repo's master because right now, there are some changes in the pull request that will undo the last two commits I've made to master. 
Sorry for the confusion but as soon as we get this sorted, I can merged in your PR.

ukblewis:
Hi @estolfo,

I've rebased. Can you please merge?


Warmest regards,

Ben

estolfo:
Hi @ukblewis 
I'm sorry for not catching this earlier, but the change to return a Symbol instead of a String from Symbol#to_bson_key is a behavior change so we can't merge it into master until we are ready to release a major version. 
That said, we have two options:
1. I already have a branch for 5.0 because I have another change requiring a major release. You could open this pull request against [5.0-dev](https://github.com/mongodb/bson-ruby/tree/5.0-dev) instead.
2. The only other Ruby object that is a valid bson key is an Integer after being converted to a String. What you could do instead of changing the #to_bson_key method on Symbol, is put a #valid_bson_key? method on each Ruby object and then call it [here](https://github.com/mongodb/bson-ruby/blob/master/lib/bson/hash.rb#L45), then expand ``rb_bson_byte_buffer_put_cstring`` to handle integers as well.

What do you think?
Emily

ukblewis:
Hi @estolfo,

No worries at all :)

I have switched the base to 5.0-dev so that it can be merged as part of 5.0 release. However, I see that there are conflicts - I will resolve them.

I think that maintaining `to_bson_key`'s functionality doesn't make sense, however, I do see that you are using the `to_s` in the Integer class for `to_bson_key`, so hopefully, when I get time, I will also write a C and Java method to enable outputting those to BSON more efficiently too and we can create the new `valid_bson_key?` method. Once we have done that, we can delete the `to_bson_key` and keep the code base clean. Does that sound sensible? What is the planned timeline for the 5.0 release?



Warmest regards,

Ben

ukblewis:
( @estolfo For your information:
Since I've built my new PR from the 5.0-dev branch merging my code into my 5.0-dev clone, I'm closing this thread )

PR #80, opened 2017-04-19
Update java extension package and byte buffer names to avoid overlap with Java driver

estolfo:

PR #83, opened 2017-05-24
Improve Ruby Symbol handling

ukblewis:


ukblewis:
New PR since it's now using branch `5.0-dev`, see PR #81 for the discussion.

estolfo:
Hi Ben,

Thank you for changing this PR to be against 5.0.
I didn't have a set date to release 5.0 because the work currently in this branch wasn't incredibly pressing but we can consider the release a higher priority now that we have your optimizations. We could release it in the next few weeks, for example.

One thing about changing put_cstring to handle Integers as well that would be a problem, is that the String representation of the Integer needs to be validated if validating_keys is true. Because having Integer keys is a much less frequent use-case than having Symbols or Strings, we could also just leave the #to_bson_key implementation on Integer as-is. We could also put the validating of the key in the c code. What do you think?

Emily

estolfo:
Hi @ukblewis 
I've pushed a commit to 5.0-dev and that handles converting integers to strings for bson keys in the extension code: https://github.com/mongodb/bson-ruby/pull/80/commits/eef5e3b3f3de5d93a42f27099f82a22109f5d9d8

Let me know what you think!
Emily

estolfo:
Hi @ukblewis
I was wondering if you saw my above comment? Let me know if you've had a chance to take a look at the commit.
Thanks!
Emily

ukblewis:
Hi @estolfo ,

Yes I did, sorry that I didn't get back to you earlier, I was really busy at work. It looks great :) what is the status with the 'validate_keys' ? Do you still want me to take a look at this?


Warmest regards,

Ben

estolfo:
Hi @ukblewis 
With the above commit, we won't have to validate the keys. It essentially applies the same behavior to an Integer as you've applied to a Symbol. In other words, an Integer is converted to a string in the putCstring/rb_bson_byte_buffer_put_cstring methods.

ukblewis:
Hi @estolfo ,

I see, I guess I misunderstood. I thought that the validations had to be run on a string instance? Did we not mention trying to do the validation in the C/Java?


Warmest regards,

Ben

estolfo:
Yes, you're right - the validation does need to be done on a string but the validation only checks that "Keys may not start with '$' or contain a '.'" (ref: https://github.com/mongodb/bson-ruby/blob/master/lib/bson/string.rb#L34 and https://github.com/mongodb/bson-ruby/blob/master/lib/bson/string.rb#L117)

An Integer cannot have a $ as the first character and wouldn't have a '.' either so we don't have to validate the result of integer.to_s

ukblewis:
@estolfo Ahhh wonderful :)

The best solutions always are the simplest. Thank you for all your help!

Have a good afternoon.

estolfo:
And thank you!

PR #88, opened 2017-12-05
Apply symbol-string key conversion to BSON::Document#dig too.

stulentsev:
Found this the hard way :(


estolfo:
thank you @stulentsev !

PR #92, opened 2018-04-20
update ruby versions in evergreen

saghm:


saghm:
I actually already pushed this separately (before I realized we hadn't updated the Ruby versons): https://github.com/mongodb/bson-ruby/commit/7b7fdaeaedfed86fbf29d77da99bf38fdecbf634

estolfo:
ah, ok great. Thanks for clarifying.

PR #95, opened 2018-07-05
Fix compile warnings

p-mongo:


behackett:
If we care about the extensions building on Windows with VC++, these changes are necessary.

PR #97, opened 2018-07-27
RUBY-1379 Test on ruby-head in evergreen

p-mongo:


.evergreen/run-tests.sh
```diff
@@ -9,33 +9,42 @@ set -o errexit  # Exit the script with error if any of the commands fail
...
-  rvm reinstall $RVM_RUBY
-fi
+  # 12.04, 14.04 and 16.04 are good
+  wget -O ruby-head.tar.bz2 http://rubies.travis-ci.org/ubuntu/`lsb_release -rs`/x86_64/ruby-head.tar.bz2
+  tar xf ruby-head.tar.bz2
+  export PATH=`pwd`/ruby-head/bin:`pwd`/ruby-head/lib/ruby/gems/2.6.0/bin:$PATH
+  ruby --version
+  ruby --version |grep dev
+  
+  #rvm reinstall $RVM_RUBY
```
saghm: If this line isn't needed anymore, we can just delete it; there's no need to save things in comments when we can go back and find them in git history if we need them.

p-mongo: It is documentation for how ruby-head is handled on rvm.

saghm: Let's just put a comment (in prose) saying what we want to document then, since it looks like we just commented it out to stop it from running

p-mongo:
@saghm Made the requested changes.

PR #99, opened 2018-08-06
RUBY-1411 Make fetch work with symbols

p-mongo:


lib/bson/document.rb
```diff
@@ -34,6 +34,25 @@ module BSON
   # @since 2.0.0
   class Document < ::Hash
 
+    # Get a value from the document for the provided key. Can use string or
+    # symbol access, but the fastest will be to always provide a key that is of
```
saghm: I'm having trouble finding the documentation for convert_key (which I assume is part of Hashie), but is this true even if we're always calling `convert_key`? I'd prefer to see some documentation stating that this is faster based on the type of the key in the hash before stating this in our docs.

p-mongo: Symbols are always converted to strings, making strings faster. The data is also stored as strings I would imagine making the comment technically accurate, though definitely confusing to a user.

saghm: Can we just change it to "using a string will be faster"?

p-mongo: That is what I did.

PR #104, opened 2018-10-19
Adding ActiveSupport::TimeWithZone patch

zachmccormick:
When using Mongoid, it properly handles the conversion to a Time object _then_ `to_bson`. Because `ActiveSupport::TimeWithZone` uses `method_missing` to respond to `to_bson` by calling `time` (and thus dropping the time zone), using `ActiveSupport::TimeWithZone` today causes incorrect timestamps to be used when using the driver directly. This patch solves it.


p-mongo:
@zachmccormick Thank you for the patch. Can you please describe the issue you were having, specifically:

- What you did
- What you expected to happen
- What actually happened

p-mongo:
https://jira.mongodb.org/browse/RUBY-1529

zachmccormick:
Yep! So if you run some code (Ruby/Mongoid) like:

```
good_time = Time.now
bad_time = Time.now.in_time_zone("Pacific Time (US & Canada)")
User.collection.find(:u_at => {:$lt => bad_time}).to_a.length # this query will be wrong

# the problem is that:
good_time.to_bson != bad_time.to_bson
```

zachmccormick:
Mongoid actually handles it correctly, so 
```
User.where(:u_at => {:$lt => bad_time}).to_a.length
```
will result in the correct query

p-mongo:
Okay I see.

p-mongo:
I outlined the next steps for this PR in https://jira.mongodb.org/browse/RUBY-1529.

PR #115, opened 2019-01-18
RUBY-1677 Fix BSON::Document#fetch method signature

fanfilmu:
JIRA link: https://jira.mongodb.org/browse/RUBY-1677

After adding the `fetch` method (in #99) it changed it's signature. This PR makes sure it's the same as Ruby's `Hash#fetch`

### DIFF
--- a/lib/bson/document.rb
+++ b/lib/bson/document.rb
@@ -37,15 +37,30 @@ module BSON
     # Get a value from the document for the provided key. Can use string or
     # symbol access, with string access being the faster of the two.
     #
+    # @overload fetch(key)
+    #   Returns a value from the hash for the given key. If the key does
+    #   not exist, raises KeyError exception.
+    #
+    # @overload fetch(key, default)
+    #   Returns a value from the hash for the given key. If the key does not
+    #   exist, returns *default*.
+    #
+    # @overload fetch(key, &block)
+    #   Returns a value from the hash for the given key. If the key does not
+    #   exist, returns the value of the block called with the key.
+    #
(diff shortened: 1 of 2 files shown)


lib/bson/document.rb
```diff
@@ -44,12 +44,15 @@ class Document < ::Hash
...
+    # @param [ Object ] default Returned value if key does not exist
+    # @yield [key] Block returning default value for given key
     #
     # @return [ Object ] The found value. Raises KeyError if none found.
     #
     # @since 4.4.0
-    def fetch(key)
-      super(convert_key(key))
+    def fetch(key, *args, &block)
+      key = convert_key(key)
```
saghm: Why do we take `*args` here? `Hash#fetch` only takes a single optional argument after the key. I think the correct signature here would be  `def fetch(key, default = nil, &block)`.

p-mongo: As written this is forward compatible should Ruby decide to add another argument to fetch.

saghm: Sure, but it also makes it show up in the documentation as taking more arguments than can be passed to it without raising an error. I don't think that we should make any method we define that wraps a standard library one accept as many arguments as possible just in case it changes at some point.

p-mongo: The behavior is also different when passing nil for default and not passing a default at all.

I updated the docstring following the suggestion in https://stackoverflow.com/questions/3934008/rdoc-darkfish-and-the-call-seq-tag#4628982.

PR #119, opened 2019-02-25
RUBY-1653 Replace BigDecimal.new with BigDecimal() 

jonekdahl:
To fix warning in Ruby 2.6:

> BigDecimal.new is deprecated; use BigDecimal() method instead

Similar to mongodb/mongoid#4616


p-mongo:
Thanks!

jonekdahl:
Hi @p-mongo, when will there be a new version of this gem with this patch included?

PR #120, opened 2019-05-24
RUBY-1958 Add GitHub repository URL to gemspec

ybiquitous:
Purpose:
- This change adds the link to the gem page on <https://rubygems.org>.
- This allows 3rd-party tools (e.g. Dependabot) to generate links to the repository, compare page, releases page, and so on.

See also:
- https://guides.rubygems.org/specification-reference/#metadata
- https://rubygems.org/gems/bson

Example:
- https://rubygems.org/gems/rspec-rails
- https://github.com/rspec/rspec-rails/blob/v3.8.2/rspec-rails.gemspec#L16-L22
image

### DIFF
--- a/bson.gemspec
+++ b/bson.gemspec
@@ -16,6 +16,10 @@ Gem::Specification.new do |s|
   s.description       = 'A full featured BSON specification implementation, in Ruby'
   s.license           = 'Apache-2.0'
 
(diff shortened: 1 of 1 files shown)


egiurleo:
Thanks for the contribution, @ybiquitous!

PR #121, opened 2019-06-25
Return the original document as the normalized document

kylemacey:
`BSON::Document` currently inherits `#to_bson_normalized_value` from `Hash` and unnecessarily casts itself into a new document. This creates some unexpected behavior:

```ruby
[1] pry(main)> require "bson"
=> true
[2] pry(main)> doc = BSON::Document.new
=> {}
[3] pry(main)> val = BSON::Document.new
=> {}
[4] pry(main)> doc["key"] = val
=> {}
[5] pry(main)> val.object_id == doc["key"].object_id
=> false
```

This new behavior doesn't cast a provided BSON::Document to a new document, saving performance and memory, while providing expected behavior.

```ruby
[4] pry(main)> doc = BSON::Document.new
=> {}
[5] pry(main)> val = BSON::Document.new
=> {}
[6] pry(main)> doc["key"] = val
=> {}
[7] pry(main)> val.object_id == doc["key"].object_id
=> true
```

### DIFF
--- a/lib/bson/document.rb
+++ b/lib/bson/document.rb
@@ -207,6 +207,17 @@ module BSON
 
     alias :update :merge!
 
+    # Returns a normalized BSON value, which is the original document.
+    #
+    # @example Return the valid BSON document
+    #   document.to_bson_normalized_value
+    #
+    # @return [ BSON::Document ] The original document.
+    #
+    def to_bson_normalized_value
+      self
+    end
+
     if instance_methods.include?(:dig)
       # Retrieves the value object corresponding to the each key objects repeatedly.
       # Will normalize symbol keys into strings.
--- a/spec/bson/document_spec.rb
+++ b/spec/bson/document_spec.rb
@@ -163,6 +163,14 @@ describe BSON::Document do
     it "sets the value" do
       expect(doc[key]).to eq(val)
     end
+
+    context "the value is a BSON::Document" do
+      let(:val) { described_class.new }
+
+      it "returns the same object" do
+        expect(doc[key].object_id).to eq(val.object_id)
+      end
+    end
   end
 
   if described_class.instance_methods.include?(:dig)
@@ -280,6 +288,13 @@ describe BSON::Document do
 
     it "returns the key/value pairs as an array" do
       expect(doc.to_a).to eq(keys.zip(vals))
+    end
+  end
+
+  describe "#to_bson_normalized_value" do
+
+    it "returns itself" do
+      expect(doc.to_bson_normalized_value.object_id).to eq(doc.object_id)
     end
   end
 


p-mongo:
Thank you for the patch.

Can you please clarify what your use case is that would benefit from this change? A recursive data structure cannot be serialized, hence I am having trouble understanding the practical utility of creating recursive data structures.

<pre>
speed% irb -Ilib -rbson
irb(main):001:0> doc = BSON::Document.new
=> {}
irb(main):002:0> val = BSON::Document.new
=> {}
irb(main):003:0> doc["val"] = doc
=> {"val"=>{...}}
irb(main):004:0> doc
=> {"val"=>{...}}
irb(main):005:0> doc['val']
=> {"val"=>{...}}
irb(main):006:0> doc['val']['val']
=> {"val"=>{...}}
irb(main):007:0> doc.to_bson
Traceback (most recent call last):
fatal (machine stack overflow in critical region)
</pre>

kylemacey:
@p-mongo thanks for the quick reply! I was using this in a Mongoid application, here's a more practical use case, where I wanted to seed a Hash (actually a BSON document) with a key.

```ruby
class Test
  include Mongoid::Document

  field :data, type: Hash

  def meta
    # I tried using a Hash first before I realized that I was interfacing with a BSON Document
    # data["_meta"] ||= {}
    data["_meta"] ||= BSON::Document.new
  end
end
```

Before the patch:

```ruby
> test = Test.new
> test.meta[:foo] = "bar" # <- This is interacting with the unserialized BSON::Document that I instantiated with
#=> "bar"
> test.meta # <- But this is returning the casted BSON::Document, a new object
#=> {} 
> test.meta[:foo] = "bar" 
#=> "bar"
> test.meta 
#=> {"foo" => "bar"}
```

With the patch

```ruby
> test = Test.new
> test.meta[:foo] = "bar"
#=> "bar"
> test.meta
#=> {"foo" => "bar"}
```

kylemacey:
I'm also realizing I screwed up my original code sample, I'll update momentarily

kylemacey:
Sorry, there wasn't supposed to be any test of recursion, I simply had a brain malfunction, which I'm sure increases the confidence in my PR 😆

p-mongo:
You are probably running into https://jira.mongodb.org/browse/MONGOID-2951 which, going by the comments in the ticket, affects various data types and changing how BSON deserialization of hashes works won't address the other data types. My current impression of this issue is it needs to be resolved within Mongoid.

kylemacey:
While that is a similar issue, and Mongoid could stand to help mitigate that for a Set or Array, I would still say that in isolation of `bson-ruby`, this is unexpected behavior:

```ruby
require 'bson'

class Test
  def data
    @data ||= BSON::Document.new
  end

  def meta
    data["_meta"] ||= BSON::Document.new
  end
end

test = Test.new 

test.meta[:foo] = "bar"
#=> "bar"

test.meta
#=> {}

test.meta[:foo] = "bar"
#=> "bar"

test.meta
#=> {"foo" => "bar"}
```

It still remains that the `BSON::Document` being passed in `#meta` is being unnecessarily cast into a new object.

p-mongo:
I agree that that behavior is a bug, however I would prefer to fix it without changing the semantics of attribute assignment with respect to copying the argument.

kylemacey:
How do you propose I change my solution? 

I think what's fictitious here is that values are being cast in `#[]=`. Where we think we're assigning a value to a key, we're not really because the value is being cast. It's hard to work around that without breaking existing behavior, so perhaps another method for setting a key value *without* casting? 

```ruby
doc = BSON::Document.new
doc.set_without_casting(:key, :value)
#=> :value
```

So then our code would look like..

```ruby
require 'bson'

class Test
  def data
    @data ||= BSON::Document.new
  end

  def meta
    if data["_meta"] 
      data["_meta"]
    else
      data.set_without_casting("_meta", BSON::Document.new)
    end
  end
end

test = Test.new 

test.meta[:foo] = "bar"
#=> "bar"

test.meta
#=> {"foo" => "bar"}
```

p-mongo:
The aspect that is unclear to me is why an assignment returns anything other than the value which was assigned (i.e. actually written, post any transformations). I haven't looked at the related code in depth to answer this question yet but this is the direction I would go in.

kylemacey:
@p-mongo That's part of the ruby spec, which is why the way `#[]=` is a little strange. You can't define a `=` method that returns anything other than what's passed in.

```ruby
[32] pry(main)> class Foo
[32] pry(main)*   def bar= baz
[32] pry(main)*     return "darth vader"
[32] pry(main)*   end
[32] pry(main)* end
=> :bar=
[33] pry(main)> f = Foo.new
=> #<Foo:0x007fbe07a31e30>
[34] pry(main)> f.bar = "luke"
=> "luke"
```

p-mongo:
I described the issue in https://jira.mongodb.org/browse/RUBY-1979. It affects other types besides Document, which is the only one that this PR discusses.

As far as I can tell the behavior is how Ruby works, and therefore not something we can change.

I added an explanation and suggested pattern which works around this issue in https://github.com/mongodb/bson-ruby/pull/141.

PR #144, opened 2019-10-30
RUBY-1977 clarify more edge cases and update tutorial


p-mongo:


p-mongo:
I will remove jruby 9.1 support in the next PR.

PR #152, opened 2019-12-18
RUBY-2046 Add BSON Binary Subtype 6 to list of binary subtypes

egiurleo:
Specification is [here](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/subtype6.rst).

This subtype is used to communicate with mongocryptd and store encrypted values in MongoDB.

(Not sure if there's anything more I need to do here.)

### DIFF
--- a/lib/bson/binary.rb
+++ b/lib/bson/binary.rb
@@ -39,6 +39,7 @@ module BSON
       :uuid_old => 3.chr,
       :uuid => 4.chr,
       :md5 => 5.chr,
+      :cyphertext => 6.chr,
       :user => 128.chr
     }.freeze
 
--- a/lib/bson/binary.rb
+++ b/lib/bson/binary.rb
@@ -31,6 +31,12 @@ module BSON
 
     # The mappings of subtypes to their single byte identifiers.
     #
+    # @note subtype 6 (ciphertext) is used for the Client-Side Encryption
+    #   feature. Data represented by this subtype is often encrypted, but
+    #   may also be plaintext. All instances of this subtype necessary for
+    #   Client-Side Encryption will be created internally by the Ruby driver.
+    #   An application should not create new BSON::Binary objects of this subtype.
+    #
     # @since 2.0.0
     SUBTYPES = {
       :generic => 0.chr,
@@ -39,7 +45,7 @@ module BSON
       :uuid_old => 3.chr,
       :uuid => 4.chr,
       :md5 => 5.chr,
-      :cyphertext => 6.chr,
+      :ciphertext => 6.chr,
       :user => 128.chr
     }.freeze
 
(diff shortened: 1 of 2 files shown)


lib/bson/binary.rb
```diff
@@ -39,6 +39,7 @@ class Binary
       :uuid_old => 3.chr,
       :uuid => 4.chr,
       :md5 => 5.chr,
+      :cyphertext => 6.chr,
```
p-mongo: "Ciphertext" is spelled with "i".

egiurleo: 😅 Turns out I don't know how to spell. Thank you.

egiurleo:
@p-mongo I think that's a really good point. I've thought of some alternatives:
1. encryption_related (mention encryption, but unlike "ciphertext," doesn't imply that the message itself is encrypted)
2. intent_to_encrypt_or_encrypted_bson (long but meaningful)

We could also name it encrypted_bson so that it matches the specification and then put a note in the documentation that explains further.

Do any of those sound better? Is there something else you'd prefer?

p-mongo:
Could you look at how the drivers that released FLE GA handle this situation?

egiurleo:
No other driver has labeled this subtype yet. I spoke to @jyemin and @kevinAlbs last night, and they pointed out that the intent to encrypt marking is an implementation detail of Mongocryptd and is not application-facing, so from the point of view of an application using the Ruby driver, BSON binary subtype 6 is always an encrypted value. I'm in favor of keeping it as "ciphertext" or some variety ("encrypted_value," "encrypted," etc). What do you think?

p-mongo:
So an application should never need to construct or work with data of this subtype?

If so keeping the label as "ciphertext" is reasonable with a comment above it stating this type is private to the driver and should not be used by the applications for either reading or writing.

egiurleo:
An application might work with this subtype because this is the type that will be stored in the database. So hypothetically, an application could insert encrypted data in the subtype 6 format, and then fetch that data with a client that can't do auto decryption, which means some of the fields in the fetched document will be BSON Binary subtype 6. The application will never have to _create_ a binary of this subtype and it will never have to interact with an intent to encrypt marking.

I'll add a note and update the spelling :)

lib/bson/binary.rb
```diff
@@ -31,6 +31,12 @@ class Binary
 
     # The mappings of subtypes to their single byte identifiers.
     #
+    # @note subtype 6 (ciphertext) is used for the Client-Side Encryption
+    #   feature. Data represented by this subtype is often encrypted, but
+    #   may also be plaintext. All instances of this subtype necessary for
+    #   Client-Side Encryption will be created internally by the Ruby driver.
+    #   An application should not create new BSON::Binary objects of this subtype.
```
egiurleo: @p-mongo Do you think this is clear?

p-mongo: Yes this is great, thank you.

PR #157, opened 2020-01-08
RUBY-2061 Update corpus spec runner for current spec guidance

p-mongo:
https://jira.mongodb.com/browse/RUBY-2061


p-mongo:
https://jira.mongodb.org/browse/RUBY-2061

p-mongo:
This is a partial implementation of the spec runner as without extended json parsing we can't round-trip *from* any of the extended json representations.

PR #160, opened 2020-01-09
RUBY-2066 Extended json serialization


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2066

ixti:
This PR breaks cases when Time has nanoseconds:

``` ruby
Time.utc(2020, 02, 02, 23, 59, 59, 999_999)
```

Due to `#to_f` not able to correctly represent nanoseconds (which is written in Ruby documentation):

> Note that IEEE 754 double is not accurate enough to represent the exact number of nanoseconds since the Epoch.
> -- https://ruby-doc.org/core-2.6.3/Time.html#method-i-to_f

That means, all Rails apps using `{ $lte: time.end_of_* }` are doomed to get absolutely wrong result, time will be serialized as beginning of next day.

p-mongo:
@ixti Thank you for the report. Can you please create an issue at https://jira.mongodb.org/browse/RUBY which includes a snippet reproducing the problem, the expected result and the actual result?

ixti:
@p-mongo Sure, I'm trying to prepare a good example.

ixti:
https://jira.mongodb.org/browse/RUBY-2145

PR #162, opened 2020-01-13
RUBY-2066 use a value that serializes the same in all drivers


p-mongo:
https://jira.mongodb.com/browse/RUBY-2066


p-mongo:
This was upstreamed as https://github.com/mongodb/specifications/pull/704.

PR #168, opened 2020-01-14
RUBY-2078 Implement Symbol and Int64 round-tripping

p-mongo:
https://jira.mongodb.com/browse/RUBY-2078


p-mongo:
https://jira.mongodb.org/browse/RUBY-2078

lib/bson/symbol.rb
```diff
@@ -89,18 +89,73 @@ def to_bson_normalized_key
...
+    # @option options [ true | false ] :relaxed Whether to produce relaxed
+    #   extended JSON representation.
+    #
+    # @return [ Hash ] The extended json representation.
+    def as_extended_json(**options)
+      { "$symbol" => to_s }
+    end
+
+    class Raw
+      def initialize(str_or_sym)
```
p-mongo: The methods added here also need docstrings. I was going to  copy them from some other class.

spec/bson/int64_spec.rb
```diff
@@ -71,6 +71,25 @@
     it_behaves_like "a bson element"
     it_behaves_like "a deserializable bson element"
 
+    context 'canonical deserialization' do
```
p-mongo: A similar test should be added for BSON::Symbol::Raw, I think it would probably have to go into symbol_spec because `from_bson`  would be called on Symbol not BSON::Symbol::Raw.

lib/bson/int64.rb
```diff
@@ -35,13 +35,20 @@ class Int64
     #
     # @param [ ByteBuffer ] buffer The byte buffer.
     #
+    # @option options [ nil | :bson ] :mode Decoding mode to use.
+    #
     # @return [ Integer ] The decoded Integer.
```
p-mongo: This return value needs to be adjusted to be Integer | Int64.

lib/bson/symbol.rb
```diff
@@ -89,18 +89,73 @@ def to_bson_normalized_key
...
+
     module ClassMethods
 
       # Deserialize a symbol from BSON.
       #
       # @param [ ByteBuffer ] buffer The byte buffer.
       #
+      # @option options [ nil | :bson ] :mode Decoding mode to use.
+      #
       # @return [ Regexp ] The decoded symbol.
```
p-mongo: This return value needs to be fixed to be Symbol | Symbol::Raw with the discussed changes.

PR #175, opened 2020-01-21
RUBY-2083 Restore legacy extended BSON spec tests

egiurleo:


spec/spec_tests/corpus_legacy_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      spec.invalid_tests.each do |test|
+
+        context("INVALID CASE: #{test.description}") do
+
+          let(:error) do
+            begin; test.reencoded_bson; false; rescue => e; e; end
+          end
+
+          it 'raises an error' do
+            skip 'This test case does not raise and error but should' unless error
```
egiurleo: This was in the original test runner. These tests do not raise errors. I'm not sure if I should do anything to change that.

p-mongo: These tests test parsing failures, and expect parses to fail. However the driver parses some of the inputs successfully, thus there is no error. I think the behavior hasn't changed in this regard singe 4.7.0.

spec/spec_tests/corpus_legacy_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+              expect(test.extjson_from_bson).to eq(test.correct_extjson)
+              expect(test.extjson_from_bson[test.test_key]).to eq(test.correct_extjson[test.test_key])
+            end
+
+            it 'decodes from extended json, then encodes the document as extended json correctly' do
+              expect(test.extjson_from_encoded_extjson).to eq(test.correct_extjson)
+              expect(test.extjson_from_encoded_extjson[test.test_key]).to eq(test.correct_extjson[test.test_key])
+            end
+
+            it 'can be converted to extended json using the #as_json method', if: spec.klass.respond_to?(:as_json) do
```
p-mongo: It does not appear that any of these blocks actually run. I replaced them with what I had in mind originally.

PR #181, opened 2020-01-22
RUBY-2081 Update ext_json to parse legacy regexes

egiurleo:


lib/bson/ext_json.rb
```diff
@@ -303,10 +307,32 @@ module ExtJSON
...
+          unless sorted_keys == %w($options $regex)
+            raise "Invalid $regex value: #{hash}"
+          end
+
+          if hash['$regex'].is_a?(Hash)
+            return {
+              '$regex' => parse_hash(hash['$regex']),
+              '$options' => hash['$options']
+            }
+          end
```
egiurleo: @p-mongo Let me know if this is the desired behavior here.

PR #174, opened 2020-01-21
RUBY-2092 Fix build on Mingw and add Windows-CI

larskanis:
Original issue is: https://github.com/oneclick/rubyinstaller2/issues/167

See commit messages for description. The CI result looks like so: https://ci.appveyor.com/project/larskanis/bson-ruby

This github project has to be registered at https://ci.appveyor.com/projects/new , so that each `git push` triggers a CI build.

PR #183, opened 2020-01-28
Add Appveyor badge icon and change icons all to SVG format

larskanis:
This is in response to https://github.com/mongodb/bson-ruby/pull/174#issuecomment-579360045

The project id "kq2b5dqv9ay132a2" is taken from RubyInstaller2 and needs to be exchanged by that of bson, to show the correct state.


p-mongo:
Thank you, https://github.com/mongodb/bson-ruby/pull/183

PR #187, opened 2020-02-07
RUBY-2122 Test Symbol de/serialization with its bson_type configured to both string and symbol


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2122

PR #189, opened 2020-02-28
RUBY-2146 Support older driver versions for BSON deserialization


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2146

PR #190, opened 2020-03-02
Fix RUBY-2145 bson 4.8.0 rounds Time during serialization to bson/extended json


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2145

PR #192, opened 2020-03-04
RUBY-2158 Return UTC times when parsing extended json


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2158

PR #195, opened 2020-04-21
Read CC environment variable

tbaederr:
To allow compilation with non-gcc compilers

For reference, this is also what e.g. the sqlite rubygem does: https://github.com/sparklemotion/sqlite3-ruby/blob/master/ext/sqlite3/extconf.rb#L7

### DIFF
--- a/ext/bson/extconf.rb
+++ b/ext/bson/extconf.rb
@@ -1,3 +1,4 @@
 require "mkmf"
 $CFLAGS << " -Wall -g -std=c99"
+RbConfig::MAKEFILE_CONFIG['CC'] = ENV['CC'] if ENV['CC']
 create_makefile("bson_native")


p-mongo:
Hi @tbaederr , thank you for the PR.

In order to accept a change like this I would like to see at least one, and ideally both, of the following:

1. Some sort of official Ruby documentation that explains the need for the change and documents the parts involved. For the PR you are proposing, this documentation would need to explain how `CC` environment variable is meant to be handled in C extensions.

2. A *thorough* description of the use case demanding this change.

Are you trying to compile bson-ruby using  a different compiler from the one Ruby was compiled with? If yes, I would like to see some references to official Ruby documentation stating that such an arrangement is supported, and how one is meant to configure their environment and/or the C extensions to make it work.

tbaederr:
Didn't know the fallback was what-ruby-was-built-with and not just gcc. Closing.

PR #201, opened 2020-06-08
Fix RUBY-2273 Regexp::Raw cannot be deserialized from YAML


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2273

PR #202, opened 2020-06-22
RUBY-2185 Fix compaction-related segfault

egiurleo:
When `GC.compact` is called while using the Ruby driver, the bson gem sometimes segfaults. This is because there are global variables in the bson C extension that are not properly marked during GC, causing their memory to sometimes be compacted and the C extension to find another object in their place, hence the segfaults.

This is fixed by calling `rb_gc_register_mark_object` and passing in the VALUEs -- this tells the C extension to mark those global variables as roots during GC, which prevents their values from being moved.

Question: should there be a test for this?

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -79,6 +79,7 @@ functions:
               export MACHINE="${MACHINE}"
               export CI=1
               export WITH_ACTIVE_SUPPORT="${WITH_ACTIVE_SUPPORT}"
+              export COMPACT="${COMPACT}"
            EOT
            # See what we've done
            cat expansion.yml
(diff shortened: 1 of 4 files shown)


p-mongo:
I would like a test.

One way of testing this is to randomly call `GC.compact` while the test suite is running. We could add an after hook (executed per test) that (globally) randomly selects a probability between, say, 100 and 1000, and after each example, with that probability, performs `GC.compact`. If this triggers the segfault in bson test suite that would be great. I would do this as a new configuration.

Additionally we could have the same configuration added to the driver test suite where there are many more tests.

p-mongo:
Looks like ruby-head on windows is still broken with https://github.com/rubygems/rubygems/issues/3690, no action needed from us I imagine on this.

PR #205, opened 2020-06-25
RUBY-2063 Changed Timestamps to allow for unsigned 64 bit integers

Neilshweky:
I did this by adding a new function to the C code called put_uint32 which allows for unsigned 32 bit integers. (You need two 32 bit integers, seconds and increment, which gives the above mentioned 64 bits)

I have not implemented the Java version yet.

### DIFF
--- a/ext/bson/init.c
+++ b/ext/bson/init.c
@@ -88,7 +88,7 @@ void Init_bson_native()
    * call-seq:
    *   buffer.get_array(**options) -> Array
    *
-   * Reads an array from the byte buffer..
+   * Reads an array from the byte buffer.
    *
    * @option options [ nil | :bson ] :mode Decoding mode to use.
    *
@@ -100,7 +100,7 @@ void Init_bson_native()
   
   /*
    * call-seq:
-   *   buffer.get_uint32(buffer) -> FixNum
+   *   buffer.get_uint32(buffer) -> Fixnum
    *
    * Reads an unsigned 32 bit number from the byte buffer.
    *
--- a/ext/bson/init.c
+++ b/ext/bson/init.c
@@ -97,6 +97,7 @@ void Init_bson_native()
   rb_define_method(rb_byte_buffer_class, "get_array", rb_bson_byte_buffer_get_array, -1);
 
   rb_define_method(rb_byte_buffer_class, "get_int32", rb_bson_byte_buffer_get_int32, 0);
+  rb_define_method(rb_byte_buffer_class, "get_uint32", rb_bson_byte_buffer_get_uint32, 0);
   rb_define_method(rb_byte_buffer_class, "get_int64", rb_bson_byte_buffer_get_int64, 0);
   rb_define_method(rb_byte_buffer_class, "get_string", rb_bson_byte_buffer_get_string, 0);
 
--- a/ext/bson/init.c
+++ b/ext/bson/init.c
@@ -97,6 +97,17 @@ void Init_bson_native()
   rb_define_method(rb_byte_buffer_class, "get_array", rb_bson_byte_buffer_get_array, -1);
 
   rb_define_method(rb_byte_buffer_class, "get_int32", rb_bson_byte_buffer_get_int32, 0);
+  
+  /*
+   * call-seq:
+   *   buffer.get_uint32(buffer) -> FixNum
+   *
+   * Reads an unsigned 32 bit number from the byte buffer..
+   *
+   * @return [ Fixnum ] The unsigned 32 bits integer from the buffer
+   *
+   * @api private
+   */
   rb_define_method(rb_byte_buffer_class, "get_uint32", rb_bson_byte_buffer_get_uint32, 0);
   rb_define_method(rb_byte_buffer_class, "get_int64", rb_bson_byte_buffer_get_int64, 0);
   rb_define_method(rb_byte_buffer_class, "get_string", rb_bson_byte_buffer_get_string, 0);
--- a/ext/bson/init.c
+++ b/ext/bson/init.c
@@ -102,7 +102,7 @@ void Init_bson_native()
    * call-seq:
    *   buffer.get_uint32(buffer) -> FixNum
    *
-   * Reads an unsigned 32 bit number from the byte buffer..
+   * Reads an unsigned 32 bit number from the byte buffer.
    *
    * @return [ Fixnum ] The unsigned 32 bits integer from the buffer
    *
--- a/ext/bson/init.c
+++ b/ext/bson/init.c
@@ -204,6 +204,18 @@ void Init_bson_native()
 
   /*
    * call-seq:
+   *   buffer.put_uint32(fixnum) -> ByteBuffer
+   *
+   * Writes an unsigned 32-bit integer value to the buffer.
+   *
+   * If the argument cannot be represented in 32 bits, raises RangeError.
+   *
+   * Returns the modified +self+.
+   */
+  rb_define_method(rb_byte_buffer_class, "put_uint32", rb_bson_byte_buffer_put_uint32, 1);
+
+  /*
+   * call-seq:
    *   buffer.put_int64(fixnum) -> ByteBuffer
    *
    * Writes a 64-bit integer value to the buffer.
--- a/ext/bson/init.c
+++ b/ext/bson/init.c
@@ -211,6 +211,9 @@ void Init_bson_native()
    * If the argument cannot be represented in 32 bits, raises RangeError.
    *
    * Returns the modified +self+.
+   *
+   * @api private
+   *
    */
   rb_define_method(rb_byte_buffer_class, "put_uint32", rb_bson_byte_buffer_put_uint32, 1);
 
--- a/ext/bson/write.c
+++ b/ext/bson/write.c
@@ -396,7 +396,7 @@ VALUE rb_bson_byte_buffer_put_uint32(VALUE self, VALUE i)
   }
 
   temp = NUM2LL(i);
-  if (temp < 0 || temp >= pow(2, 32)) {
+  if (temp < 0 || temp > UINT32_MAX) {
     rb_raise(rb_eRangeError, "Number %lld is out of range [0, 2^32)", (long long)temp);
   }
 
--- a/ext/bson/write.c
+++ b/ext/bson/write.c
@@ -392,7 +392,7 @@ VALUE rb_bson_byte_buffer_put_uint32(VALUE self, VALUE i)
   uint32_t i32;
 
   if (RB_TYPE_P(i, T_FLOAT)) {
-    rb_raise(rb_eArgError, "put_uint32; incorrect type: float, expected: integer");
+    rb_raise(rb_eArgError, "put_uint32: incorrect type: float, expected: integer");
   }
 
   temp = NUM2LL(i);
--- a/ext/bson/write.c
+++ b/ext/bson/write.c
@@ -392,6 +392,9 @@ VALUE rb_bson_byte_buffer_put_uint32(VALUE self, VALUE i)
   if (RB_TYPE_P(i, T_FLOAT))
     rb_raise(rb_eArgError, "put_uint32; incorrect type: float, expected: integer");
 
+  if (NUM2LONG(i) < 0) {
+    rb_raise(rb_eRangeError, "put_uint32; inputted number cannot be less than 0");
+  }
 
   const uint32_t i32 = NUM2UINT(i);
 
--- a/ext/bson/write.c
+++ b/ext/bson/write.c
@@ -395,7 +395,7 @@ VALUE rb_bson_byte_buffer_put_uint32(VALUE self, VALUE i)
 
   int64_t temp = NUM2LL(i);
   if (temp < 0 || temp >= pow(2, 32)) {
-    rb_raise(rb_eRangeError, "Number %ld is out of range [0, 2^32)", (long)temp);
+    rb_raise(rb_eRangeError, "Number %lld is out of range [0, 2^32)", (long long)temp);
   }
 
   const uint32_t i32 = NUM2UINT(i);
--- a/ext/bson/write.c
+++ b/ext/bson/write.c
@@ -392,7 +392,7 @@ VALUE rb_bson_byte_buffer_put_uint32(VALUE self, VALUE i)
   if (RB_TYPE_P(i, T_FLOAT))
     rb_raise(rb_eArgError, "put_uint32; incorrect type: float, expected: integer");
 
-  if (NUM2LONG(i) < 0) {
+  if (NUM2LL(i) < 0) {
     rb_raise(rb_eRangeError, "put_uint32; inputted number cannot be less than 0");
   }
 
--- a/ext/bson/write.c
+++ b/ext/bson/write.c
@@ -389,11 +389,13 @@ VALUE rb_bson_byte_buffer_put_uint32(VALUE self, VALUE i)
 {
   byte_buffer_t *b;
 
-  if (RB_TYPE_P(i, T_FLOAT))
+  if (RB_TYPE_P(i, T_FLOAT)) {
     rb_raise(rb_eArgError, "put_uint32; incorrect type: float, expected: integer");
+  }
 
-  if (NUM2LL(i) < 0) {
-    rb_raise(rb_eRangeError, "put_uint32; inputted number cannot be less than 0");
+  int64_t temp = NUM2LL(i);
+  if (temp < 0 || temp >= pow(2, 32)) {
+    rb_raise(rb_eRangeError, "Number %ld is out of range [0, 2^32)", temp);
   }
 
   const uint32_t i32 = NUM2UINT(i);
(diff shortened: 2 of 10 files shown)


ext/bson/init.c
```diff
@@ -202,6 +202,18 @@ void Init_bson_native()
...
 
+  /*
+   * call-seq:
+   *   buffer.put_uint32(fixnum) -> ByteBuffer
+   *
+   * Writes an unsigned 32-bit integer value to the buffer.
+   *
+   * If the argument cannot be represented in 32 bits, raises RangeError.
+   *
+   * Returns the modified +self+.
```
p-mongo, line 213: Can you add `@api private` at the end of this docblock please.

lib/bson/timestamp.rb
```diff
@@ -110,6 +110,8 @@ def as_extended_json(**options)
     #
     # @since 2.0.0
     def initialize(seconds, increment)
+      if seconds < 0 then seconds += 2**32 end
+      if increment < 0 then increment += 2**32 end
```
p-mongo, line 114: I don't think either of these values should be negative. If they are negative in C code, they need to be fixed in C code also.

Neilshweky: do you think I should create a get_uint32 then?

p-mongo: That sounds like a reasonable approach.

ext/bson/init.c
```diff
@@ -97,6 +97,7 @@ void Init_bson_native()
   rb_define_method(rb_byte_buffer_class, "get_array", rb_bson_byte_buffer_get_array, -1);
 
   rb_define_method(rb_byte_buffer_class, "get_int32", rb_bson_byte_buffer_get_int32, 0);
+  rb_define_method(rb_byte_buffer_class, "get_uint32", rb_bson_byte_buffer_get_uint32, 0);
```
p-mongo, line 100: This should also be `@api private`.

spec/bson/byte_buffer_read_spec.rb
```diff
@@ -101,6 +101,25 @@
     end
   end
 
+  describe '#get_uint32' do
+
+    let(:buffer) do
+      described_class.new("#{4294967295.to_bson.to_s}")
```
p-mongo, line 107: I believe this can simply be `described_class.new(4294967295.to_bson)`.

Neilshweky: This sort of thing pops up in a bunch of other places, should I replace it all?

Neilshweky: I replaced what I could be replaced... Let me know what you think.

p-mongo: What happens to this test if you use 4294967294?

Neilshweky:
```1) BSON::ByteBuffer#get_uint32 gets the uint32 from the buffer
     Failure/Error: expect(int32).to eq(4294967295)
     
       expected: 4294967295
            got: 4294967294
     
       (compared using ==)
```

Neilshweky: i.e. it works if I change the expected output of the test

p-mongo: If you change input to 4294967294 do you get 4294967294 back?

spec/bson/byte_buffer_spec.rb
```diff
@@ -30,16 +30,37 @@
...
+
+          it 'returns the length of the buffer' do
+            expect(buffer.length).to eq(4)
+          end
+        end
+
+        context 'when number is not in range' do 
+          it 'raises error on out of range number' do
+            expect{ buffer.put_uint32(4294967296) }.to raise_error(RangeError)
+            expect{ buffer.put_uint32(-4294967296) }.to raise_error(RangeError)
```
p-mongo, line 57: Each of these assertions should be in its own example, and the negative test should use -1 as the value closest to the valid range.

src/main/org/bson/ByteBuf.java
```diff
@@ -306,6 +306,24 @@ public RubyFixnum getInt32() {
...
+   *
+   * @author Neil Shweky
+   * @since 2020.06.26
+   */
+  @JRubyMethod(name = "get_uint32")
+  public RubyFixnum getUInt32() {
+    ensureBsonRead();
+
+    long temp = this.buffer.getInt();
+    if (temp < 0) temp += Math.pow(2,32);
```
p-mongo, line 320: Space after comma please, and use the block syntax with {} and the statement on a separate line please.

src/main/org/bson/ByteBuf.java
```diff
@@ -488,6 +506,35 @@ public ByteBuf putInt32(ThreadContext context, IRubyObject value) {
...
+   *
+   * @param value The integer to write.
+   *
+   * @author Neil Shweky
+   * @since 2020.06.26
+   */
+  @JRubyMethod(name = "put_uint32")
+  public ByteBuf putUInt32(ThreadContext context, IRubyObject value) {
+    if (value instanceof RubyFloat) {
+      value = ((RubyFloat) value).truncate(context);
```
p-mongo, line 520: Let's add a test case for using put_uint32 with a floating-point number. Since this is a new method, we are not bound by compatibility requirements and we can raise an exception in this case. The method should raise ArgumentError when given floating-point input. If `fix2long` raises this exception, the check here is not necessary, otherwise the check should remain but instead of truncating raise the exception.

src/main/org/bson/ByteBuf.java
```diff
@@ -488,6 +506,35 @@ public ByteBuf putInt32(ThreadContext context, IRubyObject value) {
...
+    }
+    ensureBsonWrite(4);
+
+    long temp = RubyNumeric.fix2long((RubyFixnum) value);
+    
+    if (temp >= Math.pow(2, 32) || temp < 0)
+      throw getRuntime().newRangeError(format("Number %d is out of range [0, 2^32)", temp));
+    
+    if (temp > Math.pow(2,31)) temp -= Math.pow(2,32);
```
p-mongo, line 529: Shouldn't this subtract 2**31? If this issue didn't fail any tests can a test be added to detect it?

src/main/org/bson/ByteBuf.java
```diff
@@ -488,6 +506,35 @@ public ByteBuf putInt32(ThreadContext context, IRubyObject value) {
...
+
+
+
```
p-mongo, line 532: 2 of the 3 empty lines here should be deleted.

spec/bson/byte_buffer_spec.rb
Neilshweky: -1 is not out of the range... It automatically gets converted into an unsigned integer (4294967295 to be specific), that's why I put -4294967296 as the lower range.

src/main/org/bson/ByteBuf.java
Neilshweky: This should be 2^32 actually... Say the number is 4294967295 (2^32-1), this doesn't fit into a signed int, so I have to convert this to it's respective negative number, which is -1. Therefore I have to subtract by 2^32. I can add a test...

spec/bson/byte_buffer_read_spec.rb
Neilshweky: Yes, I added a test for that...

src/main/org/bson/ByteBuf.java
Neilshweky: I added error handling for both JRuby and MRI for this case

spec/bson/byte_buffer_spec.rb
p-mongo: -1 is not an unsigned integer, therefore my first thought was it should raise an error. @egiurleo thoughts?

spec/bson/byte_buffer_spec.rb
```diff
@@ -30,16 +30,62 @@
...
-      before do
-        buffer.put_int32(5)
+      context '#put_int32' do 
+        before do
+          buffer.put_int32(5)
+        end
+
+        it 'returns the length of the buffer' do
+          expect(buffer.length).to eq(4)
+        end
```
egiurleo, line 40: Correct me if I'm wrong here but this looks like a duplicate of the test you have below?

spec/bson/byte_buffer_spec.rb
egiurleo: I think I agree with Oleg here. If the user wants to store 4294967295, they should specify 4294967295 rather than -1. Of course those numbers are the same as far as the byte buffer is concerned, but to the user they are different numbers and I believe our API should take that into account.

spec/bson/byte_buffer_write_spec.rb
```diff
@@ -585,6 +585,15 @@
...
   end
 
+  describe '#put_uint32' do
+    context 'when argument is a float' do
+      it 'appends the int32 to the byte buffer' do
+        expect{ buffer.put_uint32(4.934) }.to raise_error(ArgumentError, "put_uint32; incorrect type: float, expected: integer")
+      end
+
+    end
+  end
```
egiurleo, line 595: I'd love to see some examples that test the extreme cases that should work or raise an error -- i.e. 0 and 4294967295, and -1 and 4294967296.

spec/bson/byte_buffer_spec.rb
Neilshweky: Got it!

spec/bson/byte_buffer_spec.rb
Neilshweky: The one right below it is put_uint32 (unsigned) while this one is put_int32 (signed)

spec/bson/byte_buffer_write_spec.rb
Neilshweky: I think I put those in byte_buffer_spec.rb, would you like me to transfer them?

egiurleo: Yeah I think so -- because this method is implemented in the `write` file of the C extension, I think it makes sense to do the bulk of the testing in this file. Thanks for asking :)

Neilshweky: done!

src/main/org/bson/ByteBuf.java
```diff
@@ -488,6 +506,34 @@ public ByteBuf putInt32(ThreadContext context, IRubyObject value) {
...
+  public ByteBuf putUInt32(ThreadContext context, IRubyObject value) {
+    if (value instanceof RubyFloat) {
+      throw getRuntime().newArgumentError("put_uint32; incorrect type: float, expected: integer");
+    }
+    ensureBsonWrite(4);
+
+    long temp = RubyNumeric.fix2long((RubyFixnum) value);
+    
+    if (temp >= Math.pow(2, 32) || temp < 0) {
+      throw getRuntime().newRangeError(format("Number %d is out of range [0, 2^32)", temp));
```
egiurleo, line 525: Is there a reason to provide different error messages for the java extensions and C extensions? I think that could be potentially confusing to both users and developers.

ext/bson/init.c
```diff
@@ -97,6 +97,18 @@ void Init_bson_native()
   rb_define_method(rb_byte_buffer_class, "get_array", rb_bson_byte_buffer_get_array, -1);
 
   rb_define_method(rb_byte_buffer_class, "get_int32", rb_bson_byte_buffer_get_int32, 0);
+  
+  /*
+   * call-seq:
+   *   buffer.get_uint32(buffer) -> FixNum
+   *
+   * Reads an unsigned 32 bit number from the byte buffer..
```
egiurleo, line 105: Small typo. Sorry I didn't notice earlier.

spec/bson/byte_buffer_read_spec.rb
```diff
@@ -101,10 +101,67 @@
...
+        buffer.get_uint32
+      end
+
+      it 'gets the uint32 from the buffer' do
+        expect(int32).to eq(4294967295)
+      end
+
+      it 'increments the position by 4' do
+        expect(buffer.read_position).to eq(4)
+      end
```
egiurleo, line 120: I think you want to put an end here before the next context.

ext/bson/init.c
p-mongo: Extra period at the end.

ext/bson/init.c
```diff
@@ -97,6 +97,18 @@ void Init_bson_native()
   rb_define_method(rb_byte_buffer_class, "get_array", rb_bson_byte_buffer_get_array, -1);
 
   rb_define_method(rb_byte_buffer_class, "get_int32", rb_bson_byte_buffer_get_int32, 0);
+  
+  /*
+   * call-seq:
+   *   buffer.get_uint32(buffer) -> FixNum
```
p-mongo, line 103: `Fixnum`

spec/bson/byte_buffer_read_spec.rb
```diff
@@ -101,10 +101,67 @@
...
+        expect(int32).to eq(0)
+      end
+
+      it 'increments the position by 4' do
+        expect(buffer.read_position).to eq(4)
+      end
+    end
+  end
+
+  end
```
egiurleo, line 159: And remove this straggling end :)

ext/bson/write.c
```diff
@@ -383,6 +384,35 @@ void pvt_put_int32(byte_buffer_t *b, const int32_t i)
   b->write_position += 4;
 }
 
+/* The docstring is in init.c. */
+VALUE rb_bson_byte_buffer_put_uint32(VALUE self, VALUE i)
+{
+  byte_buffer_t *b;
+
+  if (RB_TYPE_P(i, T_FLOAT)) {
+    rb_raise(rb_eArgError, "put_uint32; incorrect type: float, expected: integer");
```
p-mongo, line 393: `;` -> `:`

ext/bson/write.c
```diff
@@ -383,6 +384,35 @@ void pvt_put_int32(byte_buffer_t *b, const int32_t i)
...
+  }
+
+  int64_t temp = NUM2LL(i);
```
p-mongo, line 396: All of the declarations should be at the top of the scope for compatibility, in this case at the beginning of the function.

ext/bson/write.c
```diff
@@ -383,6 +384,35 @@ void pvt_put_int32(byte_buffer_t *b, const int32_t i)
...
+  if (temp < 0 || temp >= pow(2, 32)) {
```
p-mongo, line 397: Does `ULONG_MAX` work here instead of `pow(2, 32)`? http://www.cplusplus.com/reference/climits/

spec/bson/byte_buffer_write_spec.rb
```diff
@@ -585,6 +585,74 @@
     end
   end
 
+  describe '#put_uint32' do
+    context 'when argument is a float' do
+      it 'appends the int32 to the byte buffer' do
```
p-mongo, line 590: The description here needs to match the behavior (i.e. an exception is raised).

spec/spec_tests/data/corpus/timestamp.json
```diff
@@ -13,6 +13,11 @@
             "canonical_bson": "100000001161002A00000015CD5B0700",
             "canonical_extjson": "{\"a\" : {\"$timestamp\" : {\"t\" : 123456789, \"i\" : 42} } }",
             "degenerate_extjson": "{\"a\" : {\"$timestamp\" : {\"i\" : 42, \"t\" : 123456789} } }"
+        },
+        {
+            "description": "Timestamp with high-order bit set on both seconds and increment",
+            "canonical_bson": "10000000116100FFFFFFFFFFFFFFFF00",
+            "canonical_extjson": "{\"a\" : {\"$timestamp\" : {\"t\" : 4294967295, \"i\" :  4294967295} } }"
```
p-mongo, line 20: Can this example be changed to use 4000000000 please.

ext/bson/write.c
Neilshweky: It didn't work for me, but also the value will depend on the operating system it's running on and I think it will cause CI errors...

src/main/org/bson/ByteBuf.java
```diff
@@ -98,6 +98,11 @@
    */
   private int writePosition = 0;
 
+  /**
+   * The size of an unsigned 32-bit integer: 2^32
+   */
+  private static long UINT32_SIZE = (long)Math.pow(2, 32);
```
p-mongo, line 104: Can you please calculate the 2^32-1 and provide it here as a literal (not 2^32), and rename the variable to UINT32_MAX to match the C code.

src/main/org/bson/ByteBuf.java
```diff
@@ -101,7 +101,12 @@
   /**
    * The size of an unsigned 32-bit integer: 2^32
    */
-  private static long UINT32_SIZE = (long)Math.pow(2, 32);
+  private static long UINT32_SIZE = 4294967296L;
+
+  /**
+   * The max of an signed 32-bit integer: 2^31
+   */
+  private static long INT32_MAX = 2147483648L;
```
p-mongo, line 109: I think this should be `2147483647L`, and it seems like there needs to be a test added that would fail with `2147483648L` being used instead of `2147483647L`.

Neilshweky: the test shouldn't fail if it's 2147483648L, being that it puts an unsigned integer, 2^31 can fit there... I put a test there to verify that my code correctly converts 2^31 into something that can fit into a signed int

PR #206, opened 2020-06-27
RUBY-2294 Provide clear exception message when trying to serialize types to BSON that do not define serialization behavior


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2294

PR #208, opened 2020-07-08
RUBY-2307 Segmentation fault on s390x

junaruga:
This PR is to report the bug on s390x, bson >= 4.9.0, pointed by the JIRA ticket: [RUBY-2307](https://jira.mongodb.org/browse/RUBY-2307).

Adding some s390x cases for debugging.

Travis CI: https://travis-ci.org/github/mongodb/bson-ruby/builds/706149254

Thank you.

<hr />

Notes from @egiurleo:

I believe the issue is that the `pvt_get_string` method was using the wrong variable on big endian architectures when getting the int32 that represented the string length. It already declared a multi-purpose string length variable, so I just swapped out the little endian-specific variable for the one that works in both cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,6 @@ git:
 
 before_install:
   - gem install bundler -v '<2'
-  - |
-    if [ "${COMMIT}" != "" ]; then
-      echo "Checking out ${COMMIT} ..."
-      git checkout "${COMMIT}"
-    fi
 
 rvm:
   - 2.3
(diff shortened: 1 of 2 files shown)


egiurleo:
Hey, @junaruga! I believe I've fixed this issue. Would you mind testing it out on your project and confirming that it works for you?

junaruga:
@egiurleo Sure. Let me test your patch on my project. Thanks!

By the way, this PR's Travis passes on s390x.
https://travis-ci.org/github/mongodb/bson-ruby/jobs/715573826

.travis.yml
```diff
@@ -1,13 +1,23 @@
...
 
-env: CI="travis"
+env:
+  global:
+    - CI="travis"
+
+jobs:
+  include:
+    - arch: s390x
+      rvm: 2.7.1
```
junaruga, line 20: How about changing this part `rvm: 2.7.1` to `rvm: 2.7` to run on the latest Ruby 2.7? It's good to run on Ruby 2.7.X in the future.

egiurleo: Good call! I'll fix that.

junaruga:
> Sure. Let me test your patch on my project. Thanks!

I checked the bson with your patch on s390x in my project (Fedora project). And it works.
Thanks!

egiurleo:
Fantastic! This fix will be released in the upcoming 4.10 patch release and backported to 4.9 as well.

PR #212, opened 2020-07-16
Fix RUBY-2128 JRuby ByteBuffer advances read & write positions together

Neilshweky:
I did this by switching the mode on flipping the buffer.

### DIFF
--- a/spec/bson/hash_spec.rb
+++ b/spec/bson/hash_spec.rb
@@ -244,6 +244,16 @@ describe Hash do
           /(Hash value for key 'foo'|Value) does not define its BSON serialized type:.*HashSpecUnserializableClass/)
       end
     end
+
+    context 'when hash is empty' do 
+      let(:buffer) do 
+        {}.to_bson
+      end
+
+      it 'has length 0' do 
+        expect(Hash.from_bson(buffer)).to eq({})
+      end
+    end
   end
 
   describe '#to_bson' do
--- a/spec/bson/hash_spec.rb
+++ b/spec/bson/hash_spec.rb
@@ -250,7 +250,7 @@ describe Hash do
         {}.to_bson
       end
 
-      it 'has length 0' do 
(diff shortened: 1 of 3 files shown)


spec/bson/hash_spec.rb
```diff
@@ -244,6 +244,16 @@ class HashSpecUnserializableClass
           /(Hash value for key 'foo'|Value) does not define its BSON serialized type:.*HashSpecUnserializableClass/)
       end
     end
+
+    context 'when hash is empty' do 
+      let(:buffer) do 
+        {}.to_bson
+      end
+
+      it 'has length 0' do
```
p-mongo, line 253: This test needs a different name. The situation we are testing is reading a buffer after having written to it previously, so this needs to be in the description and not the fact that hash length is zero when the hash is empty.

src/main/org/bson/ByteBuf.java
```diff
@@ -702,12 +702,14 @@ private ByteBuf putJavaString(final String string) {
   private void ensureBsonRead() {
     if (this.mode == Mode.WRITE) {
       this.buffer.flip();
+      this.mode = Mode.READ;
```
p-mongo, line 705: What happens if one reads-writes-reads-writes, in both MRI and JRuby?

Neilshweky: In MRI, this doesn't matter since there are separate pointers for read and write positions.... In JRuby we do this whole flipping maneuver... I'm going to add some tests for this... Just wanted to see if CI passed with the current changes.

Neilshweky:
This is still in progress. Don't review just yet

Neilshweky:
This PR should be fixed by RUBY-2334.

p-mongo:
https://jira.mongodb.org/browse/RUBY-2128

p-mongo (approved):

egiurleo (approved):
Nice!

PR #214, opened 2020-08-06
RUBY-2179 Document that dates are serialized as midnight in UTC


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2179

docs/tutorials/bson-v4.txt
```diff
@@ -588,14 +588,30 @@ all time calculations using integer math, as inexactness of floating point
...
+
+
+Date Instances
+--------------
+
+BSON only supports storing the time as the number of seconds since the
+Unix epoch. Ruby's ``Date`` instances can be serialized to BSON, but when
+the BSON is deserialized the times will be returned as ``Time`` instances.
+
+When ``Date`` instances are serialized, the time value used is the midnight
```
egiurleo, line 612: I think would say "midnight" instead of "the midnight" but that's not a blocker.

p-mongo: :+1:

PR #224, opened 2020-11-03
Fix RUBY-2436 Write ByteBuffer gets length reset to 0 after to_s conversion in JRuby


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2436

egiurleo (approved):
Nice, thank you!

PR #225, opened 2020-11-06
RUBY-2397 Support parsing $uuid as extended JSON representation for subtype 4 binary

ksadoff:
I added parsing logic in `ext_json.rb` to support parsing `$uuid` keys, and I synced our corpus spec tests to include the new uuid tests.

### DIFF
--- a/lib/bson/ext_json.rb
+++ b/lib/bson/ext_json.rb
@@ -137,6 +137,7 @@ module BSON
       if hash.empty?
         return {}
       end
+
       if hash.key?('$ref')
         # Legacy dbref handling.
         # Note that according to extended json spec, only hash values (but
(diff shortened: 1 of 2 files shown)


lib/bson/ext_json.rb
```diff
@@ -219,6 +219,14 @@ module ExtJSON
             raise Error::ExtJSONParseError, "Invalid subType value in $binary: #{value}"
           end
           create_binary(encoded_value, subtype)
+
+        when '$uuid'
+          unless /\A[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}\z/.match(hash['$uuid'])
+            raise Error::ExtJSONParseError, "Invalid $uuid value: #{hash}"
```
p-mongo, line 225: The value given in the error message should be the value being validated, to be consistent with the other branches. I suggest assigning `hash[$uuid]` to a variable and using it here and in the `from_uuid` call below.

p-mongo (approved):

egiurleo (approved):

PR #227, opened 2020-12-30
RUBY-2438: to_uuid does not add the fourth dash to serialized uuid

alexbevi:


lib/bson/binary.rb
```diff
@@ -207,7 +207,8 @@ def to_uuid(representation = nil)
         if representation && representation != :standard
           raise ArgumentError, "Binary of type :uuid can only be stringified to :standard representation, requested: #{representation.inspect}"
         end
-        data.split('').map { |n| '%02x' % n.ord }.join.sub(/(.{8})(.{4})(.{4})(.{12})/, '\1-\2-\3-\4')
+
+        data.split('').map { |n| '%02x' % n.ord }.join.sub(/(.{8})(.{4})(.{4})(.{4})(.{12})/, '\1-\2-\3-\4-\5')
```
p-mongo, line 211: To guard against this type of issue in the future, can you please add `\A` and `\z` anchors to the regular expressions changed by this PR?

spec/bson/binary_uuid_spec.rb
```diff
@@ -129,12 +129,8 @@ def make_binary(uuid_hex_str, type)
...
       end
 
-      it 'decodes as java legacy' do
-        expect(binary.to_uuid(:java_legacy).gsub('-', '').upcase).not_to eq("00112233445566778899AABBCCDDEEFF")
-      end
-
-      it 'decodes as python legacy' do
-        expect(binary.to_uuid(:python_legacy).gsub('-', '').upcase).not_to eq("00112233445566778899AABBCCDDEEFF")
+      it 'expects four dashes when output as String' do
+        expect(binary.to_uuid(:csharp_legacy).scan(/(?=-)/).count).to eq(4)
```
p-mongo, line 133: Can you please change these new tests to perform equality comparisons to the expected values, rather than counting equal signs? And the removed tests should remain I expect since they don't seem to be related to the changes proposed in this PR.

alexbevi: I overlooked the previous tests were doing a `not_to eq` and should not have removed them.

lib/bson/binary.rb
alexbevi: Added the anchors

lib/bson/binary.rb
```diff
@@ -229,7 +230,7 @@ def to_uuid(representation = nil)
           hex
         else
           raise ArgumentError, "Invalid representation: #{representation}"
-        end.sub(/(.{8})(.{4})(.{4})(.{12})/, '\1-\2-\3-\4')
+        end.sub(/(.{8})(.{4})(.{4})(.{4})(.{12})/, '\1-\2-\3-\4-\5')
```
p-mongo, line 233: This should also have the anchors added.

alexbevi: Updated

p-mongo:
Thanks!

PR #228, opened 2021-01-05
RUBY-2487 Add a local implementation of BSON::Document#except to work around ActiveSupport version not working correctly for symbol keys

fcheung:
As described in https://jira.mongodb.org/projects/RUBY/issues/RUBY-2487

BSON::Document inherits `except` from Hash either when

- activesupport has added its extensions to Hash
- the ruby version is >= 3.0.0

The implementation in activesupport < 6.0 is such that keys were being
converted from symbols but this is not true of the implementation in
activesupport >= 6 or in ruby >= 3.0

Unlike `slice` this doesn't check whether the method is already defined - when I did that specs failed when run with WITH_ACTIVE_SUPPORT=1. I believe this is because active support is required after bson, so when bson/document.rb is being loaded `except` doesn't exist (yet) and active support later drops it's (bad) implementation on top.

I believe this same order loading thing affects slice - it is less obvious because ruby 2.5, 2.6 have Hash#slice builtin (so the code to override slice kicks in) but on ruby 2.4 I get a spec failure in the slice specs running `bundle exec rake spec WITH_ACTIVE_SUPPORT=1`

CI doesn't run with WITH_ACTIVE_SUPPORT=1 so the CI runs that are green below don't actually validate that the issue is fixed.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -560,10 +560,18 @@ axes:
   - id: "as"
     display_name: ActiveSupport
     values:
-      - id: "as"
-        display_name: AS
+      - id: "5.2"
+        display_name: 5.2
         variables:
-          WITH_ACTIVE_SUPPORT: true
+          WITH_ACTIVE_SUPPORT: "~> 5.2.0"
+      - id: "6.0"
+        display_name: 6.0
+        variables:
+          WITH_ACTIVE_SUPPORT: "~> 6.0.0"
+      - id: "6.1"
+        display_name: 6.1
+        variables:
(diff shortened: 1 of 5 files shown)


p-mongo (approved):

durran (approved):
LGTM

p-mongo:
Hi @fcheung , thank you for the PR.

We do actually test activesupport configurations in Evergreen which is MongoDB's CI system.

If you encountered an issue with slice can you please create a new ticket and/or a new PR and we'll be happy to look into that as well.

p-mongo:
I believe I reproduced the slice issue in https://jira.mongodb.org/browse/ruby-2491.

PR #230, opened 2021-01-12
RUBY-2491 RUBY-2504 Add slice method for all ruby versions, make it always return BSON::Document instances

ptrippett:
Backport the .slice method from ActiveSupport core ext  hash but maintain the logic to normalize the symbol keys into strings.
    
Updated inline documentation to match return values, these return values now match the return types from ActiveSupport core ext Hash

[ close RUBY-2491, RUBY-2504 ]

### DIFF
--- a/lib/bson/document.rb
+++ b/lib/bson/document.rb
@@ -285,8 +285,11 @@ module BSON
     #
     # @since 4.3.1
     def slice(*keys)
-      keys.map{|key| convert_key(key)}
-          .each_with_object(BSON::Document.new) {|key, hash| hash[key] = self[key] if has_key?(key)}
+      keys.each_with_object(self.class.new) do |key, hash|
+        if key?(key)
+          hash[key] = self[key]
+        end
+      end
     end
 
     # Returns a new document consisting of the current document minus the
--- a/lib/bson/document.rb
+++ b/lib/bson/document.rb
@@ -269,25 +269,24 @@ module BSON
       end
     end
 
-    if instance_methods.include?(:slice)
-      # Slices a document to include only the given keys.
-      # Will normalize symbol keys into strings.
-      # (this method is backported from ActiveSupport::Hash)
-      #
-      # @example Get a document/hash with only the `name` and `age` fields present
-      #   document # => { _id: <ObjectId>, :name => 'John', :age => 30, :location => 'Earth' }
-      #   document.slice(:name, 'age') # => { name: 'John', age: 30 }
-      #   document.slice('name') # => { name: 'John' }
-      #   document.slice(:foo) # => nil
-      #
-      # @param [ Array<String, Symbol> ] *keys Keys, that will be kept in the resulting document
-      #
-      # @return [ BSON::Document ] The document with only the selected keys
-      #
-      # @since 4.3.1
-      def slice(*keys)
-        super(*keys.map{|key| convert_key(key)})
-      end
+    # Slices a document to include only the given keys.
+    # Will normalize symbol keys into strings.
+    # (this method is backported from ActiveSupport::Hash)
+    #
+    # @example Get a document/hash with only the `name` and `age` fields present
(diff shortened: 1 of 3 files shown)


ptrippett:
```
$    WITH_ACTIVE_SUPPORT='~> 5.2.0' rake
[bson] Warning: No private key present, creating unsigned gem.
mkdir -p tmp/x86_64-linux/bson_native/2.4.10
cd tmp/x86_64-linux/bson_native/2.4.10
/usr/share/rvm/rubies/ruby-2.4.10/bin/ruby -I. ../../../../ext/bson/extconf.rb
[bson] Warning: No private key present, creating unsigned gem.
creating Makefile
cd -
cd tmp/x86_64-linux/bson_native/2.4.10
/usr/bin/make
compiling ../../../../ext/bson/bytebuf.c
compiling ../../../../ext/bson/endian.c
compiling ../../../../ext/bson/init.c
compiling ../../../../ext/bson/libbson-utf8.c
compiling ../../../../ext/bson/read.c
compiling ../../../../ext/bson/util.c
compiling ../../../../ext/bson/write.c
linking shared-object bson_native.so
cd -
mkdir -p tmp/x86_64-linux/stage/lib
install -c tmp/x86_64-linux/bson_native/2.4.10/bson_native.so lib/bson_native.so
cp tmp/x86_64-linux/bson_native/2.4.10/bson_native.so tmp/x86_64-linux/stage/lib/bson_native.so
/usr/share/rvm/rubies/ruby-2.4.10/bin/ruby -I/home/paul/.rvm/gems/ruby-2.4.10/gems/rspec-core-3.10.1/lib:/home/paul/.rvm/gems/ruby-2.4.10/gems/rspec-support-3.10.1/lib /home/paul/.rvm/gems/ruby-2.4.10/gems/rspec-core-3.10.1/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb
[bson] Warning: No private key present, creating unsigned gem.
 6825/6825 |======================================================================== 100 ========================================================================>| Time: 00:00:02 

Pending: (Failures listed here are expected and do not affect your suite's status)

  1) Driver BSON Corpus Legacy spec tests (top.json): Top-level document validity INVALID CASE: Stated length less than byte count, with garbage after envelope raises an error
     # This test case does not raise and error but should
     # ./spec/spec_tests/corpus_legacy_spec.rb:61


Finished in 2.68 seconds (files took 1.62 seconds to load)
6825 examples, 0 failures, 1 pending
```

p-mongo:
Hi @ptrippett , thank you for the PR.

The desired fix here would be to make slice work regardless of when ActiveSupport was loaded. I believe unconditional definition of `slice` (or unconditional definition for Ruby versions < 2.5) would do the trick.

ptrippett:
No problem, I have made the requested changes, I also aligned the documented example with what gets returned, this is now the same as what is being returned and was being returned from Active Support core ext Hash when that was used.

lib/bson/document.rb
```diff
@@ -269,25 +269,24 @@ def dig(*keys)
...
+    #   document.slice(:foo) # => {}
+    #
+    # @param [ Array<String, Symbol> ] *keys Keys, that will be kept in the resulting document
+    #
+    # @return [ BSON::Document ] The document with only the selected keys
+    #
+    # @since 4.3.1
+    def slice(*keys)
+      keys.map{|key| convert_key(key)}
+          .each_with_object(BSON::Document.new) {|key, hash| hash[key] = self[key] if has_key?(key)}
```
p-mongo, line 289: What is the purpose of has_key?

ptrippett: It ensures we only return a subset of the keys of the document and dont add new keys by adding additional params with nil values to the document through the slice method. I'll add a test case for this

p-mongo (approved):

alcaeus (approved):

PR #235, opened 2021-04-05
Fix RUBY-2572 symbolize_keys! when called on a BSON::Document produces a broken object


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/RUBY-2572

alexbevi (approved):
LGTM

PR #238, opened 2021-08-25
Draft: #as_json should be API-ready

johnnyshields:
### Work In Progress - Not Ready for Merge

#as_json should return objects for API serialization, and be different than #as_extended JSON.

The general rules for `#as_json` are (in order of precedence):

1. Map directly to a [native JSON type](https://www.json.org/json-en.html) if it is possible to do so. Examples:
   * BSON::ObjectId --> JSON String
   * BSON::Int32 / Int64 --> JSON Integer

2. Where ActiveSupport defines `#as_json` behavior for a Ruby object, follow the same. Examples:
   * BSON::Regexp::Raw --> Ruby Regex --> JSON String `"(?-mix:dsfsd|dsf)"`
   * BSON::Timestamp --> Ruby Time --> JSON String (time)
   * BSON::Decimal128 --> (similar to Ruby BigDecimal) --> JSON String `"1234"`

3. For the remaining types not mentioned above, we need to decide whether we go with:

(a) alias to `#as_extended_json`
(b) compound object (Hash)
(c) compacted (lossy) single value representation

   * lib/bson/binary.rb  --> could just return the base64, but we'd lose the `type`
   * lib/bson/code.rb --> can return code as string
   * lib/bson/code_with_scope.rb --> can return code as string but we'd lose the `scope`
   * lib/bson/db_pointer.rb --> could return just the objectId but we'd lose the `ref` (collection)
   * lib/bson/max_key.rb --> either { "$maxKey" => 1 } or `"$maxKey"`
   * lib/bson/min_key.rb --> either { "$minKey" => 1 } or `"$minKey"`

### DIFF
--- a/lib/bson/decimal128.rb
+++ b/lib/bson/decimal128.rb
@@ -55,17 +55,16 @@ module BSON
     # @since 4.2.0
     NATIVE_TYPE = BigDecimal
 
-    # Get the Decimal128 as JSON hash data.
+    # Return a string representation of the Decimal128 use in standard
+    # application-level JSON serialization. This method is intentionally
+    # different from #as_extended_json.
     #
-    # @example Get the Decimal128 as a JSON hash.
+    # @example Get the Decimal128 as a JSON-serializable object.
     #   decimal.as_json
     #
-    # @return [ Hash ] The number as a JSON hash.
-    #
-    # @since 4.2.0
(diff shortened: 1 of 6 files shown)


johnnyshields:
Closed in favor of other PRs

PR #239, opened 2021-09-03
Raise a friendlier load error when running specs

johnnyshields:
Also:
- add missing spec README, copied from Mongoid.
- in .gitmodules, use public repo reference (https) which is what Mongoid does


p-mongo:
Hi @johnnyshields , thank you for this PR.

bson-ruby does not actually need a MongoDB server, hence several of the driver readme sections do not apply. I am going to remove them.

I will also add in the exception forwarding for those people who might benefit from it (such as Ruby driver team members ;)).

johnnyshields:
@p-mongo thanks for handling this. no big issue with including exception.

p-mongo (approved):

comandeo (approved):

PR #240, opened 2021-09-03
RUBY-2655 Fix GH actions

johnnyshields:
(GH Actions is already installed in the repo)


comandeo:
https://jira.mongodb.org/browse/RUBY-2655
* Fix GH actions
* Remove references to travis and appveyor

comandeo (approved):

.github/workflows/bson-ruby.yml
```diff
@@ -20,15 +20,13 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu, macos, windows ]
-        ruby: [ 2.3, 2.4, 2.5, 2.6, 2.7, head ]
-        include:
-          - { os: windows , ruby: mingw }
```
p-mongo, line 25: Why was windows removed please?

comandeo: Stable rubies on windows are still included. I was not sure whether we support `mingw` version. Do we need it?

johnnyshields: should be mingw64, and yes, please include this! :)

johnnyshields: (It might be `x64_mingw` actually.)

johnnyshields: However this is the Ruby "platform", not the ruby "version"

comandeo: mingw is back, as well as ruby head for macos and linux

p-mongo (approved):
:+1: Thanks!

johnnyshields:
@comandeo "windows mingw" doesn't make as a ruby version... mingw is a platform like darwin is for mac. I think there should be a different way to do this (we want to use `x64_mingw`, not the 32 bit one).

Also, should check if we are testing on ARM as well for linux and ideally mac.

PR #242, opened 2021-09-04
MONGOID-5158 - Part II - #as_json support for types which should be normalized

johnnyshields:
Implement ActiveSupport-like #as_json support for:
- BSON::Decimal128
- BSON::Int32
- BSON::Int64
- BSON::Regexp
- BSON::Timestamp


johnnyshields:
@p-mongo I do not use Regexp::Raw and Timestamp in my application. I thought it would be good to have `as_json` behave consistently for these types, even if it isn't really used much in the real-world. But it's up to you what to do here.

p-mongo:
@johnnyshields We discussed this pull request with the team. The `as_extended_json` method is for internal driver use, and exists to implement the Extended JSON specification. Types other than BSON types do not need to implement this method, and it will not be used by the driver on objects that are not of BSON types. In order to not create possible future compatibility issues, we will keep `as_extended_json` method implemented only on the classes that require it.

johnnyshields:
@p-mongo will any of these types be exposed as types in Mongoid? (e.g. BSON::Decimal128)? If so, I strongly urge you to to consider adding `as_json` for them as I have here. If not, then no need.

PR #244, opened 2021-09-04
 Float: Refactor #as_extended_json to get rid of nested ifs

johnnyshields:
Makes method more readable.


lib/bson/float.rb
```diff
@@ -57,37 +57,26 @@ def to_bson(buffer = ByteBuffer.new, validating_keys = Config.validating_keys?)
...
-              if abs > 1e15
-                '%.17g' % to_s
+      value = if infinite? == 1
+                'Infinity'
+              elsif infinite? == -1
+                '-Infinity'
+              elsif nan?
+                'NaN'
+              elsif options[:mode] == :relaxed || options[:mode] == :legacy
+                self
```
p-mongo, line 67: @johnnyshields I believe all tests are failing because this line wraps numeric literals into `$numberDouble` which shouldn't be happening.

johnnyshields: Fixed, thanks!

comandeo (approved):

PR #245, opened 2021-09-04
MONGOID-5158 - Part III - Add #as_extended_json handling to Date and DateTime

johnnyshields:
This PR fixes #as_extended_json for Date and DateTime, which were both missing it.

Note that DateTime converts to Time (`#to_time`) to keep the code simple.

There is no change to the Time class implementation, however I've cleaned up the specs a bit to match the other two classes.

### DIFF
--- a/lib/bson/date.rb
+++ b/lib/bson/date.rb
@@ -54,6 +54,22 @@ module BSON
     def bson_type
       ::Time::BSON_TYPE
     end
+
+    # Converts this object to a representation directly serializable to
+    # Extended JSON (https://github.com/mongodb/specifications/blob/master/source/extended-json.rst).
+    #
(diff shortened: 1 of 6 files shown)


johnnyshields:
@p-mongo your review here and on #242 would appreciated--would be good to wrap up these BSON PRs.

p-mongo:
Hi @johnnyshields ,

We extracted the docstring fix into https://github.com/mongodb/bson-ruby/pull/248, thank you for that.

The `as_extended_json` methods shouldn't actually be defined on DateTime and Date classes - these methods are used internally by the driver for extended json serialization and this operation only needs to be defined on Time, as per the extended json specification. Based on this we'll close this PR.

PR #253, opened 2021-11-16
freeze string literals

ahorek:
1/ freezing numeric constants has no effect
2/ regexes are already frozen
3/ freeze strings via frozen_string_literal pragma instead of freezing them explicitly

### DIFF
--- a/lib/bson.rb
+++ b/lib/bson.rb
@@ -44,12 +44,12 @@ module BSON
   # Constant for bson types that don't actually serialize a value.
   #
   # @since 2.0.0
-  NO_VALUE = -("".b)
+  NO_VALUE = ::String.new("", encoding: BINARY).freeze
 
   # Constant for a null byte (0x00).
   #
   # @since 2.0.0
-  NULL_BYTE = -(0.chr.b)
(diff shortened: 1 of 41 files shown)


p-mongo (approved):

lib/bson.rb
```diff
@@ -38,22 +39,22 @@ def self.ObjectId(string)
...
   #
   # @since 2.0.0
-  BINARY = "BINARY".freeze
+  BINARY = "BINARY"
 
   # Constant for bson types that don't actually serialize a value.
   #
   # @since 2.0.0
-  NO_VALUE = "".force_encoding(BINARY).freeze
+  NO_VALUE = -("".b)
```
comandeo, line 47: Are there any performance benefits in replacing `"".force_encoding(BINARY)` with `-("".b)`? The former is way more readable to my taste.

ahorek: yes, it's faster, but it really doesn't matter for constants.

since the gem does support Ruby 2.5+, this ```String.new(encoding: BINARY)``` should be even better, what do you think?

the previous ```"".force_encoding(BINARY)``` won't work anymore, because "" is now frozen and force_encoding is a destructive operation

comandeo (approved):

PR #255, opened 2021-12-03
RUBY-2734 Prohibit embedded null bytes in strings used as BSON::Document hash keys, regexps and regexp options

Neilshweky:
After adding the tests I noticed that some of them failed, so I went ahead and attempted to implement my own solution.
Before this patch, the following case:
```
{
  "description" : "Null byte in document key",
  "string" : "\{\"a\\u0000\": 1 }"
}
```
Was being parsed as:
```
{ "a\u0000" => 1 } 
```
which is not working as intended. 

We expect it to raise an exception if there's a null byte in the key and this patch fixes that.

### DIFF
--- a/lib/bson/error.rb
+++ b/lib/bson/error.rb
@@ -25,9 +25,9 @@ module BSON
     end
 
     # Exception raised when there is an invalid argument passed into the
-    # constructor of regexp object. This includes when the argument contians
+    # constructor of regexp object. This includes when the argument contains
     # a null byte.
-    class RegexpArgumentError < Error
+    class InvalidRegexpPattern < Error
     end
   end
 end
--- a/lib/bson/error.rb
+++ b/lib/bson/error.rb
@@ -23,5 +23,11 @@ module BSON
     # itself to BSON.
     class UnserializableClass < Error
     end
+
+    # Exception raised when there is an invalid argument passed into the
+    # constructor of regexp object. This includes when the argument contians
+    # a null byte.
+    class RegexpArgumentError < Error
+    end
   end
 end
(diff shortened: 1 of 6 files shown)


lib/bson/ext_json.rb
```diff
@@ -362,6 +362,9 @@ module ExtJSON
 
     module_function def map_hash(hash, **options)
       ::Hash[hash.map do |key, value|
+        if key.include? NULL_BYTE
+          raise Error::ExtJSONParseError, "Hash key cannot contian a null byte: #{key}"
```
p-mongo, line 366: Typo: contain

lib/bson/regexp.rb
```diff
@@ -154,6 +154,12 @@ def compile
       #
       # @since 3.0.0
       def initialize(pattern, options = '')
+        if pattern.include? NULL_BYTE
+          raise Error::ExtJSONParseError, "Regex pattern cannot contian a null byte: #{key}"
```
p-mongo, line 158: The exception class here and on the line below should be `ArgumentError` or  a new class added under `BSON::Error`. The extended json parse error isn't applicable here because this method isn't parsing extended json.

p-mongo: I think "regexp" will also be more appropriate than "regex" here since the input is Ruby variables and the Ruby class is called "regexp".

lib/bson/error.rb
```diff
@@ -23,5 +23,11 @@ class BSONDecodeError < Error
     # itself to BSON.
     class UnserializableClass < Error
     end
+
+    # Exception raised when there is an invalid argument passed into the
+    # constructor of regexp object. This includes when the argument contians
```
p-mongo, line 28: contains

lib/bson/error.rb
```diff
@@ -23,5 +23,11 @@ class BSONDecodeError < Error
...
+    # a null byte.
+    class RegexpArgumentError < Error
```
p-mongo, line 30: How about `InvalidRegexpPattern`?

p-mongo:
@comandeo ready for you

comandeo (approved):
👍

PR #256, opened 2021-12-15
RUBY-2852 fix error where driver fails on integer key

Neilshweky:
Evergreen patch for Ruby Driver with this branch of BSON ruby: https://spruce.mongodb.com/version/61ba2a57a4cf473d0db59180/tasks

The JRuby one still fails but I think it's not getting the right BSON ruby. I think that failure will be gone once this is merged into master

### DIFF
--- a/lib/bson/ext_json.rb
+++ b/lib/bson/ext_json.rb
@@ -363,7 +363,7 @@ module BSON
 
     module_function def map_hash(hash, **options)
       ::Hash[hash.map do |key, value|
-        if key.to_s.include? NULL_BYTE
+        if (key.is_a?(String) || key.is_a?(Symbol)) && key.to_s.include?(NULL_BYTE)
           raise Error::ExtJSONParseError, "Hash key cannot contain a null byte: #{key}"
         end
         [key, parse_obj(value, **options)]
--- a/lib/bson/ext_json.rb
+++ b/lib/bson/ext_json.rb
@@ -363,7 +363,7 @@ module BSON
 
     module_function def map_hash(hash, **options)
       ::Hash[hash.map do |key, value|
-        if key.include? NULL_BYTE
+        if key.to_s.include?(NULL_BYTE)
           raise Error::ExtJSONParseError, "Hash key cannot contain a null byte: #{key}"
         end
         [key, parse_obj(value, **options)]
--- a/lib/bson/ext_json.rb
+++ b/lib/bson/ext_json.rb
@@ -363,7 +363,7 @@ module BSON
 
     module_function def map_hash(hash, **options)
       ::Hash[hash.map do |key, value|
-        if key.to_s.include?(NULL_BYTE)
+        if key.to_s.include? NULL_BYTE
           raise Error::ExtJSONParseError, "Hash key cannot contain a null byte: #{key}"
         end
         [key, parse_obj(value, **options)]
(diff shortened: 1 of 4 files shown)


comandeo (approved):
Is it possible to add a test for the situation we had?

Neilshweky:
@comandeo  done!

comandeo (approved):

lib/bson/ext_json.rb
```diff
@@ -363,7 +363,7 @@ module ExtJSON
 
     module_function def map_hash(hash, **options)
       ::Hash[hash.map do |key, value|
-        if key.include? NULL_BYTE
+        if key.to_s.include? NULL_BYTE
```
p-mongo, line 366: If I am reading the diff correctly, key can be of any type. If so the behavior should be:

- for strings, use `include?` directly as before
- for symbols, use `to_s.include?` as proposed now
- for everything else, do nothing
When the key is of an unhandled type I don't think it should be validated here. If there is a concern that embedded nulls would show up elsewhere, I suggest adding a validation at a point where we actually deserialize or serialize the key to a string.

lib/bson/regexp.rb
```diff
@@ -156,9 +156,9 @@ def compile
       # @since 3.0.0
       def initialize(pattern, options = '')
         if pattern.include? NULL_BYTE
-          raise Error::InvalidRegexpPattern, "Regexp pattern cannot contain a null byte: #{key}"
-        elsif options.is_a?(String) && options.include?(NULL_BYTE)
-          raise Error::InvalidRegexpPattern, "Regexp options cannot contain a null byte: #{key}"
+          raise Error::InvalidRegexpPattern, "Regexp pattern cannot contain a null byte: #{pattern}"
+        elsif options.to_s.include? NULL_BYTE
```
p-mongo, line 160: Same as in the above hunk, given that options can be an integer:

- if `options` is a string, use `include?`
- if `options` is a symbol, use `to_s.include?`
- if `options` is an integer, do nothing
- otherwise raise an exception that the options are of an invalid type

p-mongo:
Dmitry pointed out that to_s is not needed on symbols as of Ruby 2.5, therefore please just use `include?` but only on strings and symbols and the other types either do nothing or raise an exception.

Neilshweky:
> Dmitry pointed out that to_s is not needed on symbols as of Ruby 2.5, therefore please just use `include?` but only on strings and symbols and the other types either do nothing or raise an exception.
```
neil.shweky@neil bson-ruby % ruby -v 
ruby 2.7.5p203 (2021-11-24 revision f69aeb8314) [x86_64-darwin20]
neil.shweky@neil bson-ruby % irb
irb(main):001:0> :my_sym.include?("sym")
Traceback (most recent call last):
        4: from /Users/neil.shweky/.rbenv/versions/2.7.5/bin/irb:23:in `<main>'
        3: from /Users/neil.shweky/.rbenv/versions/2.7.5/bin/irb:23:in `load'
        2: from /Users/neil.shweky/.rbenv/versions/2.7.5/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        1: from (irb):1
NoMethodError (undefined method `include?' for :my_sym:Symbol)
```
I don't think that's true... Anyway, `to_s` doesn't do anything on a string, so I can just do `key.to_s` for both Strings and Symbols

lib/bson/ext_json.rb
Neilshweky: done

lib/bson/regexp.rb
Neilshweky: done

comandeo (approved):

p-mongo (approved):

PR #257, opened 2021-12-20
RUBY-2090 RUBY-2638 RUBY-2639 Move driver DBRef class to bson-ruby

Neilshweky:
RUBY-2090 move the class over to bson-ruby
RUBY-2638 don't drop additional field during serialization, fixed as a result of subclassing hash
RUBY-2639 make dbref subclass of Hash


lib/bson/dbref.rb
```diff
@@ -0,0 +1,130 @@
...
+# distributed under the License is distributed on an 'AS IS' BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module BSON
+
+  # Represents a DBRef document in the database.
+  #
+  # @since 2.1.0
```
p-mongo, line 22: The @sinces aren't correct and need to be removed.

lib/bson/dbref.rb
```diff
@@ -0,0 +1,130 @@
...
+    # Get the DBRef as a JSON document
+    #
+    # @example Get the DBRef as a JSON hash.
+    #   dbref.as_json
+    #
+    # @return [ Hash ] The max key as a JSON hash.
+    #
+    # @since 2.1.0
+    def as_json(*args)
+      {}.merge(self)
```
p-mongo, line 65: This should use `update` rather than `merge`.

lib/bson/dbref.rb
```diff
@@ -0,0 +1,130 @@
...
+      #
+      # @param [ String ] buffer The bson representing a hash.
+      #
+      # @return [ Hash, DBRef ] The decoded hash or DBRef.
+      #
+      # @see http://bsonspec.org/#/specification
+      #
+      # @since 2.0.0
+      def from_bson(buffer, **options)
+        # bson-ruby 4.8.0 changes #from_bson API to take **options.
```
p-mongo, line 112: Since this code is now in bson-ruby, does this conditional need to remain here?

p-mongo (commented):
Are there any tests in bson corpus that bson-ruby is lacking that verify dbref operation with extra keys?

lib/bson/dbref.rb
Neilshweky: done

lib/bson/dbref.rb
Neilshweky: Done. Out of curiosity, why?

lib/bson/dbref.rb
Neilshweky: Probably not, since this is for older json-ruby's. Removed and no failures. Done.

Neilshweky:
> Are there any tests in bson corpus that bson-ruby is lacking that verify dbref operation with extra keys?

There are extra tests, but it seems like we fail those tests. I am investigating.

lib/bson/dbref.rb
p-mongo: `merge` is essentially equivalent to `dup.update` - it duplicates the receiver then merges the arguments. Since the receiver is created right here (it's the empty hash) it does not need to be duplicated.

p-mongo (approved):

p-mongo:
Feel free to merge this once Evergreen is mostly green.

PR #258, opened 2021-12-21
RUBY-2855 RUBY-2637 Fix dbref spec tests

Neilshweky:
RUBY-2637 make sure dbrefs can't be created without $id (two line fix)


lib/bson/ext_json.rb
```diff
@@ -139,7 +139,7 @@ module ExtJSON
         return {}
       end
 
-      if hash.key?('$ref')
+      if hash['$ref']&.is_a?(String) && hash.key?('$id')
```
Neilshweky, line 142: For context, from the dbref spec: 
> Assert that each document is not decoded to a DBRef model. In the context of implicit decoding, the document MUST be decoded like any other embedded document. In the context of explicit decoding, the DBRef decoding method MUST raise an error.

Neilshweky: I understand implicit decoding to be this block of code, and explicit decoding to be the changes made in the constructor for dbref.

spec/spec_tests/data/corpus/dbref.json
```diff
@@ -26,6 +26,21 @@
...
+        },
+        {
+            "description": "Sub-document resembles DBRef but $ref is not a string",
+            "canonical_bson": "2c000000036462726566002000000010247265660001000000072469640058921b3e6e32ab156a22b59e0000",
+            "canonical_extjson": "{\"dbref\": {\"$ref\": {\"$numberInt\": \"1\"}, \"$id\": {\"$oid\": \"58921b3e6e32ab156a22b59e\"}}}"
+        },
+        {
+            "description": "Sub-document resembles DBRef but $db is not a string",
+            "canonical_bson": "4000000003646272656600340000000224726566000b000000636f6c6c656374696f6e00072469640058921b3e6e32ab156a22b59e1024646200010000000000",
+            "canonical_extjson": "{\"dbref\": {\"$ref\": \"collection\", \"$id\": {\"$oid\": \"58921b3e6e32ab156a22b59e\"}, \"$db\": {\"$numberInt\": \"1\"}}}"
```
Neilshweky, line 43: Can't get the following test to pass: 
```
{
            "description": "DBRef with additional dollar-prefixed and dotted fields",
            "canonical_bson": "48000000036462726566003c0000000224726566000b000000636f6c6c656374696f6e00072469640058921b3e6e32ab156a22b59e10612e62000100000010246300010000000000",
            "canonical_extjson": "{\"dbref\": {\"$ref\": \"collection\", \"$id\": {\"$oid\": \"58921b3e6e32ab156a22b59e\"}, \"a.b\": {\"$numberInt\": \"1\"}, \"$c\": {\"$numberInt\": \"1\"}}}"
        },
```
Left it out for now

Neilshweky: done!

lib/bson/ext_json.rb
Neilshweky: After speaking with Oleg, this is the correct understanding... Resolving.

ext/bson/read.c
```diff
@@ -76,7 +76,6 @@ VALUE pvt_read_field(byte_buffer_t *b, VALUE rb_buffer, uint8_t type, int argc,
     case BSON_TYPE_STRING: return pvt_get_string(b, "String");
     case BSON_TYPE_SYMBOL: return pvt_get_symbol(b, rb_buffer, argc, argv);
     case BSON_TYPE_ARRAY: return rb_bson_byte_buffer_get_array(argc, argv, rb_buffer);
-    case BSON_TYPE_DOCUMENT: return rb_bson_byte_buffer_get_hash(argc, argv, rb_buffer);
```
Neilshweky, line 79: Instead of automatically doing `get_hash` here, we should look at the class of the function and call the internal ruby `from_bson`.

spec/bson/dbref_spec.rb
```diff
@@ -202,17 +230,95 @@
...
+
+        it 'should not raise' do
+          expect do
+            buffer
+          end.to_not raise_error
+        end
+
+        it 'has the correct class' do
+          expect(decoded['dbref']).to be_a described_class
+          expect(decoded['dbref']['dbref']).to be_a described_class
```
Neilshweky, line 302: This fails, because it is actually of type `BSON::Document`.

The reason this happens is because in `from_bson` we call the `BSON::DBRef` constructor, which just validates the dbref and calls the `BSON::Document` constructor (super). The DBRef constructor does not further parse for more DBRefs (and I'm not sure it should?)

p-mongo (approved):

Neilshweky:
@p you need the dbref from_bson method because otherwise, in jruby, we call `cls.from_bson` (`cls` being `DBRef`), which then calls the Hash's from_bson and there's no possibility for the instantiation of a DBRef.

PR #259, opened 2021-12-22
RUBY-2856 update bson-ruby docs to include dbrefs

Neilshweky:
How do I see what this looks like?

### DIFF
--- a/docs/tutorials/bson-v4.txt
+++ b/docs/tutorials/bson-v4.txt
@@ -405,6 +405,38 @@ Represents a string of JavaScript code with a hash of values.
 
   BSON::CodeWithScope.new("this.value = age;", age: 5)
 
+``BSON::DBRef``
+------------------
+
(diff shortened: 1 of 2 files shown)


p-mongo:
We don't currently have a convenient way to look at bson-ruby docs locally. For driver and Mongoid we've implemented a simple local builder, e.g. described in https://github.com/mongodb/mongo-ruby-driver/blob/master/docs/README.md. bson-ruby docs are normally rendered as part of driver documentation but chances are there is a low-effort way to port the driver renderer to bson-ruby.

p-mongo:
I added a local builder in https://github.com/mongodb/bson-ruby/pull/260.

p-mongo (approved):
The version number is increased when we actually do the release, so that the only tree marked as 4.13.0 is the actual 4.13.0 release to avoid confusion.

PR #261, opened 2021-12-23
RUBY-2857 Reorder DBRef fields to have $ref/$id/$db first

Neilshweky:


lib/bson/dbref.rb
```diff
@@ -84,7 +84,7 @@ def initialize(hash)
         end
       end
 
-      super(hash)
+      super(reorder_fields(hash))
```
p-mongo, line 88: This needs to be done before the validation so that symbol keys are accepted (e.g. `:$ref`)

lib/bson/dbref.rb
```diff
@@ -122,6 +122,26 @@ def from_bson(buffer, **options)
...
+    # Reorder the fields of the given Hash to have $ref first, $id second,
+    # and $db third. The rest of the fields in the hash can come in any
+    # order after that.
+    #
+    # @param [ Hash ] hash The input hash. Must be a valid dbref.
+    #
+    # @return [ Hash ] The hash with it's fields reordered.
+    def reorder_fields(hash)
+      reordered = {}
+      reordered['$ref'] = hash.delete('$ref')
```
p-mongo, line 138: This mutates the input which is not good, easiest way to fix this is to create a `BSON::Document` out of the `hash` which would also provide indifferent access for validation in the constructor.

lib/bson/dbref.rb
Neilshweky: done

lib/bson/dbref.rb
Neilshweky: done

p-mongo (approved):

PR #263, opened 2021-12-28
RUBY-2859 Add GH Actions tests covering Ruby 3.1

johnnyshields:


johnnyshields:
This is basically ok to merge. Looks like Ruby 3.1 not out on Windows yet.

p-mongo:
Hi @johnnyshields , thank you for the PR. This will need ruby 3.1 built in our toolchain first which I have worked on today and once that PR is merged, there may be further changes need to the other repos/shared tooling because the toolchain structure has changed a while ago.

comandeo:
I think enabling 3.1 for GitHub actions is a great idea. We will have some coverage "for free" before we make necessary changes to the toolchain.

comandeo (approved):

p-mongo (approved):

PR #265, opened 2021-12-30
add support for serializing and deserializing BigDecimals

Neilshweky:


lib/bson/float.rb
```diff
@@ -16,7 +16,7 @@
 module BSON
 
   # Injects behaviour for encoding and decoding floating point values
-  # to and from # raw bytes as specified by the BSON spec.
+  # to and from raw bytes as specified by the BSON spec.
```
Neilshweky, line 19: 9 year old typo.

docs/tutorials/bson-v4.txt
```diff
@@ -222,8 +222,8 @@ Supported Classes
 
 Core Ruby classes that have representations in the BSON specification and
 will have a ``to_bson`` method defined for them are: ``Object``, ``Array``,
-``FalseClass``, ``Float``, ``Hash``, ``Integer``, ``NilClass``, ``Regexp``,
-``String``, ``Symbol`` (deprecated), ``Time``, ``TrueClass``.
+``FalseClass``, ``Float``, ``Hash``, ``Integer``, ``BigDecimal``, ``NilClass``,
+``Regexp``, ``String``, ``Symbol`` (deprecated), ``Time``, ``TrueClass``.
```
Neilshweky, line 226: It seems like there might not be such a big addition to the docs in terms of supporting **existing** Ruby classes. Is there more that needs to be done here?

PR #266, opened 2021-12-30
RUBY-2860 Implement BigDecimal#to_bson

Neilshweky:


lib/bson/float.rb
```diff
@@ -16,7 +16,7 @@
 module BSON
 
   # Injects behaviour for encoding and decoding floating point values
-  # to and from # raw bytes as specified by the BSON spec.
+  # to and from raw bytes as specified by the BSON spec.
```
Neilshweky, line 19: 9 year old typo.

lib/bson/big_decimal.rb
```diff
@@ -0,0 +1,59 @@
...
+      # @return [ BigDecimal ] The decimal object.
+      def from_bson(buffer, **options)
+        Decimal128.from_bson(buffer, **options).to_big_decimal
+      end
+    end
+  end
+
+  # Enrich the core BigDecimal class with this module.
+  #
+  # @since 2.0.0
```
p-mongo, line 56: This needs to be taken out.

lib/bson/big_decimal.rb
```diff
@@ -0,0 +1,59 @@
...
+    def to_bson(buffer = ByteBuffer.new, validating_keys = Config.validating_keys?)
+      BSON::Decimal128.new(to_s).to_bson(buffer, validating_keys)
+    end
+
+    module ClassMethods
+
+      # Deserialize the BigDecimal from raw BSON bytes.
+      #
+      # @example Get the BigDecimal from BSON.
+      #   BigDecimal(1).from_bson(bson)
```
p-mongo, line 41: This doesn't look right.

spec/bson/big_decimal_spec.rb
```diff
@@ -0,0 +1,317 @@
...
+        expect do
+          BigDecimal(argument).to_bson
+        end.to raise_error(BSON::Decimal128::InvalidRange)
+      end
+    end
+
+    context "when passing a number with too much precision for Decimal128" do
+      let(:argument) { "1.000000000000000000000000000000000000000000000000001" }
+
+      # TODO: Change error based on Oleg's ticket
```
p-mongo, line 221: This is https://jira.mongodb.org/browse/RUBY-2863.

spec/bson/big_decimal_spec.rb
```diff
@@ -0,0 +1,317 @@
...
+
+      let(:decimal128) do
+        BSON::Decimal128.new(argument)
+      end
+
+      let(:from_bson) do
+        BigDecimal.from_bson(decimal128.to_bson)
+      end
+
+      let(:expected_from_bson) do
```
p-mongo, line 30: Usually when we use the word "expected" in a test, it refers to a literal value (e.g.  a string that is immediately specified) rather than a calculated value. "Expected" values should generally be clear as to what they are and are easy to verify for correctness.

In these tests, the variables marked "expected" are actually calculated values and it isn't obvious what they are. I suggest using a different term for them, for example you could say that this is a "decimal128 decoded value" or something similar.

lib/bson/big_decimal.rb
Neilshweky: done.

spec/bson/big_decimal_spec.rb
Neilshweky: done

spec/bson/big_decimal_spec.rb
Neilshweky: done

lib/bson/big_decimal.rb
Neilshweky: done

spec/bson/big_decimal_spec.rb
```diff
@@ -0,0 +1,317 @@
...
+        expect do
+          BigDecimal(argument).to_bson
+        end.to raise_error(BSON::Decimal128::InvalidRange)
+      end
+    end
+
+    context "when passing a number with too much precision for Decimal128" do
+      let(:argument) { "1.000000000000000000000000000000000000000000000000001" }
+
+      # TODO: RUBY-2683 update this error
```
p-mongo, line 221: You should be able to do this now after rebasing on master.

Neilshweky: done

p-mongo (approved):
:+1:

PR #273, opened 2022-01-04
RUBY-2864 Fix serialization of BigDecimal in Hash

Neilshweky:


p-mongo (approved):
Can you please add a roundtrip test as well which serializes to bson, deserializes and asserts that the result is a BSON::Decimal128 with the correct numeric value.

comandeo (approved):

PR #284, opened 2022-04-21
MONGOID-5304: Add Rubocop static code analyzer

johnnyshields:
This PR adds a static code analyzer (Rubocop) to the project, as required by MongoDB's [Technical and Organizational Security Measures](https://www.mongodb.com/technical-and-organizational-security-measures).

It is integrated with Github CI. To add it to Evergreen, simply run this shell command (only needs to be run once for the entire suite; not dependent on Ruby version, etc.) I've done a draft of Evergreen integration in this PR, but it may not work as I cannot run Evergreen myself.

```
bundle exec rake rubocop
```

Refer to the related [Mongoid PR](https://github.com/mongodb/mongoid/pull/5243) for further explanation.


johnnyshields:
Ready to merge, please check Evergreen setup works.

p-mongo (changes requested):
Hi @johnnyshields ,

As per https://jira.mongodb.org/browse/MONGOID-5304, if you would like to adjust this PR to only enable non-style/formatting checks, we will review it.

johnnyshields:
@p-mongo this is updated as per your comments in https://jira.mongodb.org/browse/MONGOID-5304

p-mongo (changes requested):
This is almost working but in evergreen the Ruby installation is under our tree and is checked by rubocop, it needs to be reconfigured to only explicitly check `lib`.

johnnyshields:
@p-mongo I can't see what Evergreen is doing, but you should be able to set:

```
AllCops:
  exclude:
    - "ruby/dir/**/*"
```

p-mongo:
I could, but I would like `lib` to be included instead.

johnnyshields:
@p-mongo is this done? Please kindly merge and then I'll port this to Mongoid and Ruby Driver libs.

johnnyshields:
@p-mongo I've updated to only include `lib` dir on a white-list basis. Please merge.

.rubocop.yml
```diff
@@ -0,0 +1,29 @@
+AllCops:
+  TargetRubyVersion: 2.5
+  NewCops: disable
+  Include:
+    - 'lib/**/*'
```
p-mongo, line 5: This needs to be the top-level `lib` only, not any of the `lib` subdirs in the current directory:

```
[2022/06/27 02:13:57.329] Running tests
[2022/06/27 02:13:57.329] Inspecting 144 files
[2022/06/27 02:13:57.329] ....................................................................................C................................................C..........
[2022/06/27 02:13:57.329] Offenses:
[2022/06/27 02:13:57.329] rubies/ruby-3.1.0/lib/ruby/gems/3.1.0/gems/ruby-prof-1.4.3/bin/ruby-prof:279:9: C: Security/Eval: The use of eval is a serious security risk.
[2022/06/27 02:13:57.329]         eval(exec)
[2022/06/27 02:13:57.329]         ^^^^
[2022/06/27 02:13:57.329] rubies/ruby-3.1/lib/ruby/gems/3.1.0/gems/ruby-prof-1.4.3/bin/ruby-prof:279:9: C: Security/Eval: The use of eval is a serious security risk.
[2022/06/27 02:13:57.329]         eval(exec)
[2022/06/27 02:13:57.329]         ^^^^
[2022/06/27 02:13:57.329] 144 files inspected, 2 offenses detected
```

johnnyshields: Hmm... this failure looks Evergreen specific. You can see the Github actions build is passing here: https://github.com/mongodb/bson-ruby/runs/7051219018?check_suite_focus=true

Since I can't access Evergreen, I think you're going to have to debug this. Possibly Evergreen is not running Rubocop from the project root dir. There are a number of things to try, some ideas:
- check Rubocop is run from project root
- try `Include:` arg as `/path/from/systemroot/bson-ruby/lib/**/*`
etc.

p-mongo: We have the Ruby installation together with gems under the project root in evergreen.

> The syntax I'm using here Include: - lib/**/* is standard as per docs and means "from project root".

`lib/**/*` in shell means `./lib/**/*`. In Rubocop this apparently means `./**/lib/**/*`. After a quick look at the configuration page I couldn't find where `Include` is described, but the configuration in this PR needs to match the shell behavior of the glob as I just described it.

johnnyshields: Again, Rubocop is working correctly on my local machine, and the specs are passing on GH Actions.

I have no way of seeing and knowing what Evergreen is doing.

johnnyshields: As a shot in the dark, let's try `/lib/**/*` instead of `lib/**/*`

p-mongo: Please find out whether `/lib/**/*` actually traverses from filesystem root and report your findings.

johnnyshields: @p-mongo I have done this already. Here's the evidence:

```
/workspace/rubocop-test$ ruby -e "puts Dir.glob('**/*')"
foo
foo/lib
foo/lib/test.rb
foo/lib/test1
foo/lib/test1/test.rb
lib
lib/test.rb
lib/test1
lib/test1/test.rb

# this is using the exact rubocop.yml in this PR
/workspace/rubocop-test$ rubocop -L
lib/test.rb
lib/test1/test.rb

# note there is also a /lib dir in my drive root
```

Please kindly run Evergreen as a next step, thanks!

p-mongo: This is the same line as it was before, no?

.rubocop.yml
```diff
@@ -0,0 +1,32 @@
...
+  Exclude:
+    - 'spec/**/*'
+    - 'vendor/**/*'
```
p-mongo, line 8: These exclusions aren't needed either, because only `lib` should be consulted per the inclusion above.

.rubocop.yml
johnnyshields: Yes. As I have stated, this line `lib/**/*` is working correctly on:
1. bson-ruby repo on my local machine
2. the test case above, which demonstrates it is only reading `lib` in root dir under normal circumstances.
3. in Github Actions, see green checkmark below

The only thing that is failing is Evergreen. In order to debug this further, running `rubocop -L` will provide a clear indication of where Rubocop is actually scanning on Evergreen, so I can fix it.

(You are making the assumption that `Include: "lib/**/*"` is causing Rubocop to scan lib folders in nested directories, however, it may just be coincidental, because most Ruby files are found in `lib` dirs somewhere.)

.rubocop.yml
johnnyshields: Removed.

.rubocop.yml
```diff
@@ -0,0 +1,30 @@
+AllCops:
+  TargetRubyVersion: 2.5
+  NewCops: disable
+  Exclude:
+    # exclude everything except lib dir
+    - !ruby/regexp /^(?!lib\/)/
```
p-mongo, line 6: Creative. So what is this value documented to be exactly?

johnnyshields: Ah sadly that Regexp didn't work. So I got more creative 😂 

Check the latest, I'm 95% confident this will work on Evergreen as well.

p-mongo:
If there isn't a normal way to tell rubocop to only check files in top-level `lib` subdir, please file this as an enhancement request with them and/or contribute the functionality to rubocop. The various workarounds I'm seeing in this PR, entertaining as they are, are not appropriate for using, of all things, a code quality tool.

johnnyshields:
Closing; replaced by https://github.com/mongodb/bson-ruby/pull/303

PR #285, opened 2022-07-11
RUBY-3047 fix README links

Neilshweky:


README.md
```diff
@@ -16,13 +16,13 @@ Documentation
 -------------
 
 Current documentation can be found
-[here](http://docs.mongodb.org/ecosystem/tutorial/ruby-bson-tutorial/#ruby-bson-tutorial).
+[here](https://www.mongodb.com/docs/ruby-driver/current/bson-tutorials/).
```
Neilshweky, line 19: I decided to go to the general ruby page instead of the v4 tutorial, so we don't have to change it every time a new bson-ruby version comes out.

PR #288, opened 2022-08-02
RUBY-2777 remove support for integer options for Regexp::Raw constructor

Neilshweky:
Since master is 5.0, I think we can remove support for integer options in Regexp::Raw constructor. This allows me to fix the bson corpus tests.

### DIFF
--- a/lib/bson/regexp.rb
+++ b/lib/bson/regexp.rb
@@ -147,11 +147,8 @@ module BSON
       #   Raw.new(pattern, options)
       #
       # @param [ String ] pattern The regular expression pattern.
-      # @param [ String, Integer ] options The options.
-      #
-      # @note The ability to specify options as an Integer is deprecated.
-      #  Please specify options as a String. The ability to pass options as
-      #  as Integer will be removed in version 5.0.0.
+      # @param [ String ] options The options.
+
       #
       # @since 3.0.0
       def initialize(pattern, options = '')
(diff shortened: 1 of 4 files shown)


lib/bson/regexp.rb
```diff
@@ -147,11 +147,8 @@ def compile
       #   Raw.new(pattern, options)
       #
       # @param [ String ] pattern The regular expression pattern.
-      # @param [ String, Integer ] options The options.
-      #
-      # @note The ability to specify options as an Integer is deprecated.
-      #  Please specify options as a String. The ability to pass options as
-      #  as Integer will be removed in version 5.0.0.
+      # @param [ String ] options The options.
+
```
Neilshweky, line 151:
```suggestion
```

lib/bson/regexp.rb
```diff
@@ -161,8 +157,8 @@ def initialize(pattern, options = '')
           if options.to_s.include?(NULL_BYTE)
             raise Error::InvalidRegexpPattern, "Regexp options cannot contain a null byte: #{options}"
           end
-        elsif !options.is_a?(Integer)
-          raise ArgumentError, "Regexp options must be a String, Symbol, or Integer"
+        else
+          raise ArgumentError, "Regexp options must be a String or Symbol"
```
p-mongo, line 161: Do you think we should disallow symbols for options? I can make a follow-up ticket if so.

p-mongo (approved):

lib/bson/regexp.rb
Neilshweky: I personally don't have a problem keeping symbol options... If you have a specific reason then I wouldn't mind dropping it but otherwise we can keep it.

p-mongo: :+1: That's fine.

PR #291, opened 2022-09-03
RUBY-3137 Use SecureRandom to initialize ObjectId counter

midnight-wonderer:
Hello,
I am auditing some MongoDB Ruby code and found that `ObjectId` uses `Kernel.rand` to initialize the counter.
While `Kernel.rand` differs from implementation to implementation, it is likely seeded by at least the PID of the Ruby process.

In the `ObjectId` use case, since we already used `PID` in the tuple, we want to make sure other values in the tuple are independent from the `PID` as much as possible.

I know it might not have a concrete downside in practice, but just to be on the safe side, I would recommend initializing the counter with `SecureRandom` instead; what do you say?

### DIFF
--- a/lib/bson/object_id.rb
+++ b/lib/bson/object_id.rb
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 require "digest/md5"
+require "securerandom"
 require "socket"
 require "thread"
 
@@ -351,7 +352,7 @@ module BSON
       #
       # @since 2.0.0
       def initialize
-        @counter = rand(0x1000000)
+        @counter = ::SecureRandom.rand(0x1000000)
         @machine_id = Digest::MD5.digest(Socket.gethostname).unpack1("N")
         @mutex = Mutex.new
       end


Neilshweky:
I have no objections, @p-mongo @comandeo?

p-mongo:
@midnight-wonderer Can you please clarify what your requirements are.

> In the ObjectId use case, since we already used PID in the tuple, we want to make sure other values in the tuple are independent from the PID as much as possible.


Can you please state what your requirements are and why you are using ObjectId for them. The behavior of ObjectId as it is used by MongoDB drivers is covered by the specification [here](https://github.com/mongodb/specifications/blob/master/source/objectid.rst#random-value), in particular the random number is not required to be cryptographically secure.

>  it is likely seeded by at least the PID of the Ruby process.

Can you please reference your source of this information?

behackett:
We do a similar thing in PyMongo: https://github.com/mongodb/mongo-python-driver/blob/master/bson/objectid.py#L51

midnight-wonderer:
@p-mongo 
First and foremost, the spec says,

> If possible, use a PRNG with OS supplied entropy that SHOULD NOT block to wait for more entropy to become available.

;for Ruby this is the `SecureRandom`.  

To add more details
MRI uses multiple sources of seed for `Kernel.rand`.
Reference:
https://github.com/ruby/ruby/blob/55c771/random.c#L655
This may or may not include `urandom` on *nix depending on how you compile it.

The objective of the spec is to lower the probability of collision on ObjectID, given that the collision on a 16-bit machine identifier should not be that rare. For Ruby's truncated `MD5(hostname)`, all it takes is people to name the same hostname on different servers. And when two identical server restarts, it's not that unusual to have Ruby with the same PID.

Even if these two events occur, the last safeguard is the counter with a randomized init value. Now the quality of randomization matters. Even if the randomization doesn't need to be cryptographically secure, the orthogonal to other values in the tuple is a desirable property.

If `Kernel.rand` was seeded with time, for example, you'll end up with something deterministic. Not independent between each process.

That part about PID has no reference, but the seed for `Kernel.rand` has pretty much the same requirement as MongoDB.
(No need for cryptographically secure randomization)

I wouldn't surprised if some implementor chooses to seed it with time / MachineID / ProcessID; this is undesirable because it will make the random number not orthogonal to other values in ObjectID.

midnight-wonderer:
Add:
Here https://github.com/ruby/ruby/blob/55c771/random.c#L680, the PID is in the seed as well, at least for MRI.

p-mongo:
Thank you for the explanation @midnight-wonderer . I agree that this PR is more aligned with the spec requirements than our present implementation, at the cost of possibly being slower.

p-mongo (approved):

Neilshweky (approved):

comandeo (approved):

comandeo:
@midnight-wonderer Thank you very much for your contribution!

PR #293, opened 2022-09-08
RUBY-3109 remove all references to dots/dollars

Neilshweky:
Some followup questions about this: 
- do we need to deprecate this in 4.15(16?)
- do we want to leave this functionality for the future?
- If not, can we remove the BSON::Config module?

Also in this PR, renamed DBRef argument error.


p-mongo:
There are many conflicting files in this PR?

ext/bson/read.c
```diff
@@ -143,7 +143,7 @@ VALUE rb_bson_byte_buffer_get_bytes(VALUE self, VALUE i)
 }
 
 VALUE pvt_get_boolean(byte_buffer_t *b){
-  VALUE result;
+  VALUE result = 0;
```
p-mongo, line 146: I don't think you can do this. If you want to silence the compiler warning, you can assign `Qnil` to `result` immediately before the `pvt_raise_decode_error` call in that branch only.

p-mongo: And add a comment that the assignment is to pacify the compiler.

Neilshweky:
> There are many conflicting files in this PR?

ooh I just merged 1806, that must be why... I will rebase.

Neilshweky:
@p-mongo  conflicts are gone!

p-mongo:
This change should be announced for 5.0. We don't currently have release notes in bson-ruby, so for now the note can be added to https://jira.mongodb.org/browse/RUBY-1604 as a simple list of changes.

If you like you can add a deprecation to 4.x. My guess is the change will affect virtually no applications. If it's easy to add the deprecation, I'd say go for it, if it's a meaningful amount of work this can be skipped in my opinion.


> do we want to leave this functionality for the future?

My guess is it won't be used in any visible future. My vote is no.

> If not, can we remove the BSON::Config module?

Good question. It might be useful but having it be there and be empty is somewhat silly. @comandeo ?

ext/bson/read.c
Neilshweky, line 146:
```suggestion
  VALUE result;
```

Neilshweky: I just got rid of it.

lib/bson/error/invalid_dbref_argument.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module BSON
+  class Error
+    # Raised when trying to create a BSON::DBRef from an object that is an invalid DBRef.
+    #
+    # @api private
```
p-mongo, line 7: Exception classes are not private, only their constructors are.

lib/bson/regexp.rb
```diff
@@ -196,15 +195,14 @@ def respond_to?(method, include_private = false)
...
       #
       # @return [ BSON::ByteBuffer ] The buffer with the encoded object.
       #
       # @see http://bsonspec.org/#/specification
       #
       # @since 4.2.0
-      def to_bson(buffer = ByteBuffer.new, validating_keys = Config.validating_keys?)
-        return compile.to_bson(buffer, validating_keys) if options.is_a?(Integer)
+      def to_bson(buffer = ByteBuffer.new)
+        return compile.to_bson(buffer) if options.is_a?(Integer)
```
p-mongo, line 205: Didn't you remove integer options?

p-mongo (approved):

lib/bson/error/invalid_dbref_argument.rb
Neilshweky: done

lib/bson/regexp.rb
Neilshweky: You're right, let me open a followup ticket.

lib/bson/error/invalid_dbref_argument.rb
p-mongo: There are 2 or 3 other classes that this applies to.

Neilshweky: done

PR #295, opened 2022-09-22
RUBY-2846 Add _bson_to_i method to ActiveSupport::TimeWithZone class

Neilshweky:


p-mongo:
This should not be necessary since https://github.com/jruby/jruby/issues/6104 is merged into JRuby 9.3 and Mongoid 8.0+ requires 9.3.

Neilshweky:
> This should not be necessary since https://github.com/jruby/jruby/issues/6104 is merged into JRuby 9.3 and Mongoid 8.0+ requires 9.3.

There is definitely work to be done in Mongoid wrt this though, right? Removing _bson_to_i and just calling to_i if we're 9.3+?

lib/bson/time_with_zone.rb
```diff
@@ -46,6 +46,16 @@ def to_bson(buffer = ByteBuffer.new)
     def bson_type
       ::Time::BSON_TYPE
     end
+
+    def _bson_to_i
```
p-mongo, line 50: Can you add `@api private` here please.

p-mongo (approved):
I changed my mind, since we don't detect jruby versions let's support correct behavior on 9.2.

When either bson-ruby is impossible to use on 9.2 or it's very clear to users this shouldn't be done, all _bson_to_i code should be removed.

lib/bson/time_with_zone.rb
Neilshweky, line 50:
```suggestion
    # @api private
    def _bson_to_i
```

Neilshweky: done

PR #296, opened 2022-09-29
RUBY-2952 fix BSON::Decimal128 to compare with other types

Neilshweky:


Neilshweky:
I have implemented option 1 but it's pretty ugly...

Neilshweky:
TODO: add test cases that evaluate to false

Neilshweky:
RUBY-2952

PR #308, opened 2023-06-22
RUBY-2056 Allow all subtype values in BSON::Binary

jamis:
`BSON::Binary` only allows subtypes to be specified as a symbol, which makes it impossible to specify any but one of the subtypes in the user-defined range (0x80-0xFF). This PR changes BSON::Binary so that the type parameter can still be given as a symbol, with the existing behavior, but also as an integer (for specifying the literal subtype value directly) or as a string (interpreted as either the symbolic type identifier, or as a single-byte integer subtype).


lib/bson/binary.rb
```diff
@@ -338,13 +347,57 @@ def self.from_uuid(uuid, representation = nil)
...
+    #
+    # @return [ Symbol ] the symbolic type corresponding to the argument.
     #
     # @raise [ BSON::Error::InvalidBinaryType ] The the type is invalid.
     #
     # @since 2.0.0
     def validate_type!(type)
-      raise BSON::Error::InvalidBinaryType.new(type) unless SUBTYPES.has_key?(type)
+      case type
+      when Integer then validate_integer_type!(type)
```
comandeo, line 359: It looks like the method `validate_type!` actually mutates the internal state of the instance by setting `@raw_type` instance variable. It took me some time to figure it out; however, I am not sure whether this is a blocker.

comandeo (approved):

lib/bson/binary.rb
jamis: It does; however, `validate_type!` is only called from `initialize`, so its less about mutating state, and more about initializing state. The `@raw_type` variable represents the actual binary encoding of the type that will be used to represent it; it is generally the same as the type itself, but for user-defined encodings (0x80-0xFF), it can be different, which is why we need to differentiate them.

PR #311, opened 2023-06-23
RUBY-1410 Implement ObjectID spec

jamis:
This implements the ObjectID spec (here: https://github.com/mongodb/specifications/blob/master/source/objectid.rst#specification), which changes how the object ID is generated.

ref: https://jira.mongodb.org/browse/RUBY-1410


comandeo (approved):

midnight-wonderer:
@jamis 
Hey, I was late to the party.  
Didn't know you guys reimplement the whole thing in C.

I have a question regarding f6abb714e94fb686c643ab0c26417636f136778d.  
How do I ensure that the extension is compiled with `arc4random`?  
Do I need `libbsd-dev` on my system (Linux)?

jamis:
Hey @midnight-wonderer. I'm actually not sure what might need to happen to find that function on a non-BSD system. Currently, the `extconf.rb` file is set to look only in the standard header files; if the file exists in a different header file, the call to `have_func` in `extconf.rb` might need to be adjusted to check for specific header files. Could possibly be done via environment variables, too? If you're up to experimenting with that, and have some success getting it to work on (e.g.) Linux, please do let us know what needs to change.

midnight-wonderer:
I tried, these are what I found out so far:
* Installing `libbsd-dev` does not work; `mkmf` doesn't detect the header.
* Surprisingly, the new version of Debian (dubbed: Bookworm) provides `arc4random` in their standard libs, and `mkmf` pick up the function just fine.
* The internal API of Ruby has a header that provides secure random with robust fallbacks; check this out: https://github.com/ruby/ruby/blob/7380c7/internal/random.h
* Normally, I write native extensions using `libffi`, so the binding will work across all Rubies without much effort. I don't know how to write Ruby in C; so, not sure if you can use `internal/random.h`.

PR #314, opened 2023-07-19
RUBY-3305 enable detections of arc4random from libbsd

midnight-wonderer:
A little change to enable `arc4random` on Linux when `libbsd` header exists.  
This change is not necessary if you can integrate `internal/random.h` from MRI as discussed in #311; that API is better and more robust.

With this change, to enable secure-randomized counter initialization for ObjectID, just make sure to install `libbsd-dev` on distros derived from Debian.

### DIFF
--- a/ext/bson/extconf.rb
+++ b/ext/bson/extconf.rb
@@ -4,6 +4,7 @@
 require 'mkmf'
 
 $CFLAGS << ' -Wall -g -std=c99'
+have_library 'bsd'
 have_func 'arc4random'
 
 create_makefile('bson_native')


jamis:
Excellent! I'll add a ticket in Jira for this and we'll triage it on Monday. I love how straightforward it was to make this work for you; I'd hoped it would be. Also, excellent suggestion about using Ruby's random utility for this; I'll add a second ticket for that so we can discuss it as a team next week.

Thank you!

jamis:
Thanks again for this PR; ultimately, we've opted for a more general solution that leans on Ruby's `SecureRandom` module (see #318).

PR #319, opened 2023-11-30
Add support for legacy deserialization

jamis:
Mongoid specs, for instance, rely on BSON::Binary defining only two instance variables, and build testing data around that assumption.


jamis:
ruby 2.7 has some failures in evergreen when COMPACT is enabled, but I've confirmed locally that these failures happen even in master. I suspect a memory glitch in ruby 2.7, as the same specs work fine on later versions of ruby.

PR #324, opened 2024-02-13
final tweaks for BSON 5 release

jamis:
Final tweaks for BSON 5 release -- had to adjust JDK versions and minimum ruby versions and the JRuby version used to bundle the java gem. Also tweaked the release script to add a "dry run" mode.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -6,11 +6,25 @@ NAME=bson
 RELEASE_NAME=bson-ruby-release
 VERSION_REQUIRE=bson/version
 VERSION_CONSTANT_NAME=BSON::VERSION
+CMD=echo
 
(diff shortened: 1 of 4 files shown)


release.sh
```diff
@@ -6,12 +6,26 @@ NAME=bson
...
 VERSION_REQUIRE=bson/version
 VERSION_CONSTANT_NAME=BSON::VERSION
+CMD=echo
 
 if ! test -f gem-private_key.pem; then
   echo "gem-private_key.pem missing - cannot release" 1>&2
   exit 1
 fi
 
+if test -z "$PRODUCTION_RELEASE"; then
```
comandeo, line 16: 👍

comandeo (approved):

PR #327, opened 2024-04-05
RUBY-3435 Don't overflow the 4-bytes allotted to an ObjectID's timestamp portion

jamis:
This PR ensures that the timestamp is (1) treated as an unsigned integer, to allow the largest possible value there, and (2) truncated to 32 bits (4 bytes) if it ever grows larger than that. This avoids a RangeError that occurs when trying to convert a Bignum to a 32-bit integer internally.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -29,6 +29,7 @@ timeout:
       script: |
         ls -la
 
+
 # -----------------------------------------------
 # .evergreen/config/functions.yml.erb
 # -----------------------------------------------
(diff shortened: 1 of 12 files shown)


johnnyshields:
I've run this in my test suite and confirmed that it fixes the bug. Thank you!

comandeo-mongo (approved):

johnnyshields:
@jamis this is running in production for us now FYI 🎉

jamis:
Wonderful, thanks for the confirmation, @johnnyshields. We'll get it officially released soon.

PR #330, opened 2024-06-25
RUBY-3501 SSDLC for bson-ruby

jamis:
The general implementation is the same as that for the driver and for Mongoid, but because we produce two deliverables (a gem for MRI, and a gem for JRuby), the release process is split into two jobs: one with a matrix to build each gem, and upload the resulting artifact, and another to download the generated artifacts and complete the publication of the gems and other deliverables.


.github/workflows/release.yml
```diff
@@ -0,0 +1,174 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [ '3.2', jruby ]
+    steps:
+      - name: Check out the repository
+        uses: mongodb-labs/drivers-github-tools/secure-checkout@v2
+        with:
+          app_id: ${{ vars.APP_ID }}
+          private_key: ${{ secrets.APP_PRIVATE_KEY }}
```
alcaeus, line 61: No objection to leaving this here, but strictly speaking this doesn't need to run in the `release` environment and use `secure-checkout`.

alcaeus (approved):
Workflows LGTM. I'll defer to @comandeo-mongo for reviewing the ruby-specific changes.

comandeo-mongo (approved):

PR #335, opened 2024-09-01
Use correct regexp lowercase \z terminator

johnnyshields:
In `VALID_DECIMAL128_STRING_REGEX`, a [recent commit](https://github.com/mongodb/bson-ruby/commit/80be3c09665297a55850bf588f0767ac4ee57708#diff-08ce2abf8e4ad4dedd852e5af813144345de9d8c4b9ae052e0204103033947b5) used `\Z` (uppercase). Probably `\z` (lowercase) is the correct one. The current regex will allow `"21.23\n"` as a valid value.

```
$ Matches the end of a line.

\z Matches the end of the string.

\Z Matches the end of the string unless the string ends with a ``\n'', in which case it matches just before the ``\n''.
```

https://ruby-doc.com/docs/ProgrammingRuby/html/language.html#UJ

### DIFF
--- a/lib/bson/decimal128/builder.rb
+++ b/lib/bson/decimal128/builder.rb
@@ -156,7 +156,7 @@ module BSON
         # @return [ Regex ] The regex for a valid decimal128 string.
         #
         # @since 4.2.0
-        VALID_DECIMAL128_STRING_REGEX = /\A[\-\+]?(\d+(\.\d*)?|\.\d+)(E[\-\+]?\d+)?\Z/i
+        VALID_DECIMAL128_STRING_REGEX = /\A[\-\+]?(\d+(\.\d*)?|\.\d+)(E[\-\+]?\d+)?\z/i
 
         # Initialize the FromString Builder object.
         #


jamis:
Hey @johnnyshields -- thanks for the PR. Is the concern that a string with a newline matches, and you believe it shouldn't?

Looking at the original regex (prior to that commit you mentioned), it used `/^...$/` anchors, and also matches the newline:

```ruby
"21.23\n" =~ /^[\-\+]?(\d+(\.\d*)?|\.\d+)(E[\-\+]?\d+)?$/i # => 0
```

The `\Z` regex preserves this behavior, whereas (as you noted) the lowercase `\z` does not. It should be noted that `Float("21.23\n")` also succeeds, so allowing a trailing newline has some precedent in Ruby itself.

If I'm misunderstanding your concern, please let me know. Otherwise, I'm inclined to let the current implementation stand.

_It should be noted in passing that the regex was changed at the recommendation of the CodeQL static analyzer, which offered `\Z` as the recommended alternative to `$`. I had no compelling reason to prefer one over the other, so opted to appease CodeQL._

johnnyshields:
If being permissive of the trailing newline is intentional, then fine to keep it. (In general `\z` is preferable to `\Z`; in the vast majority of regexp use cases one does not intend to allow trailing newlines.)

PR #338, opened 2024-10-11
RUBY-3533 Make BSON::Binary objects comparable

jamis:
Because BSON::Binary instances represent binary data, it ought to be possible to compare their raw binary contents and decide which one would sort before or after another. This PR adds an implementation of `#<=>` to BSON::Binary. As long as both objects have the same binary type, this will compare the raw contents of the two objects, allowing homogeneous lists of BSON::Binary instances to be (e.g.) sorted.

### DIFF
--- a/lib/bson/binary.rb
+++ b/lib/bson/binary.rb
@@ -24,6 +24,7 @@ module BSON
   # @since 2.0.0
   class Binary
     include JSON
+    include Comparable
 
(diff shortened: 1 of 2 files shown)


comandeo-mongo (approved):

adam-hampton-sp:
Thank you for the new feature!

PR #340, opened 2024-11-08
TEST: explicitly require stringio

mtasaka:
Since spec/spec_helper.rb uses StringIO class,
make this file explicitly require stringio.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -23,6 +23,7 @@ $LOAD_PATH.unshift(File.dirname(__FILE__))
 $LOAD_PATH.unshift(File.join(File.dirname(__FILE__), "..", "lib"))
 
 require "ostruct"
+require "stringio"
 require "bson"
 require "json"
 require "rspec"


mtasaka:
Testing with
* ruby 3.4.0dev (2024-11-07 master df3395f2e3) +PRISM [x86_64-linux] https://github.com/ruby/ruby/commit/df3395f2e301e7739dda5b7455e8c6b1c25b334d

Executing bson-ruby head 
https://github.com/mongodb/bson-ruby/commit/e1f5d8fafcf31fd17c307af883580d2be8d512e2
rspec testsuite **with ~/.spec removed** generates the following error:

```
$ env RUBYLIB=$(pwd)/lib:$(pwd)/ext/bson/ rspec

An error occurred while loading ./spec/bson/array_spec.rb.
Failure/Error:
  class AlternateIO < StringIO
  
    # Read a byte from the stream.
    #
    # @returns [ String ] A String representation of the next byte.
    def readbyte
      super.chr
    end

NameError:
  uninitialized constant StringIO
# ./spec/spec_helper.rb:68:in '<top (required)>'
# ./spec/bson/array_spec.rb:16:in '<top (required)>'


Finished in 0.00055 seconds (files took 2.75 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples
```

With ~/.spec restored, the following error does not happen, perhaps because loading rspec format style somewhere loads stringio internally in rspec. But anyway I think explicitly adding stringio dependency inside bson-ruby is desirable.

comandeo-mongo (approved):

comandeo-mongo:
@mtasaka Thank you very much for your contribution!

PR #341, opened 2024-11-14
RUBY-3557 Mark SecureRandom as Global

zacheryph:
SecureRandom needs to be marked global so that it cannot be moved during GC compaction. When this happens the `pvt_SecureRandom` may reference a different object, or nothing at all causing a segfault.

I am not quite sure how a test would be written for this since it'd have to be run on its own, and it segfaults the process so the suite wouldn't run. This can be reproduced (on master), and shown resolved (on this branch) with the following:

```
rake compile
bundle console

irb> GC.verify_compaction_references(expand_heap: true, toward: :empty);
irb> BSON::ObjectId.new
```


comandeo-mongo (approved):
@zacheryph This is fantastic, thank you very much for your contribution!

PR #353, opened 2025-05-15
Add support for jruby 10

donoghuc:
As a consumer of the mongo gem I want to run using jruby 10. 

This commit starts this work, before I get too deep in it I want to make sure i'm not duplicating effort or if this is even something that would get any buy-in. 

I noticed the first issue eaa3a1d544d8e49fda58c942abda14d73db3a0da attempting to even load the mongo gem. I attempted to add jruby 10 cells in CI to see other places that need update 63bdd199bf3efb30dc0417a8ce082c003cf45887. 

Thanks for looking :)

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -296,6 +296,10 @@ axes:
         display_name: ruby-2.7
         variables:
            RVM_RUBY: ruby-2.7
+      - id: jruby-10.0
+        display_name: jruby-10.0
+        variables:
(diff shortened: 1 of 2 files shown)


alexbevi:
Hey @donoghuc, our team had https://jira.mongodb.org/browse/RUBY-3651 on our plan for this quarter to review but haven't kicked that off yet.

No duplication of effort at this point, and we'd be happy to get another set of eyes on it

PR #360, opened 2025-06-17
Update Rakefile for correct build step

jamis:
The build step for the Rakefile was not being honored by the new release process, which was causing the Java version of the bson gem to not include a pre-built jar file. I've updated the new release process to use a rake task instead, and updated the rake task to conform to the expected interface.


alexbevi (approved):
lgtm

PR #372, opened 2026-06-03
RUBY-3894 Fix heap buffer overflow in put_string and related methods

jamis:
## Description

- `RSTRING_LEN` returns `long`, but five sites in `ext/bson/write.c` stored the result in `int32_t`, silently truncating strings ≥ 2³¹ bytes to a negative value. That value was then passed to `rb_bson_utf8_validate` as `size_t`, wrapping to near `UINT64_MAX` and driving reads far past the heap allocation (confirmed heap-buffer-overflow with AddressSanitizer).
- Fixed `pvt_bson_encode_to_utf8`, `rb_bson_byte_buffer_put_string`, `rb_bson_byte_buffer_put_cstring`, `rb_bson_byte_buffer_put_symbol`, and `pvt_put_bson_key` by changing the length variable to `long` and adding an explicit `> INT32_MAX` bounds check before narrowing back to `int32_t` for the BSON wire format. Strings exceeding the limit now raise `ArgumentError`.
- Regression tests added in `spec/bson/string_length_overflow_spec.rb`, gated behind `STRESS=1` (each example requires ~2 GB of free memory).

### DIFF
--- a/ext/bson/write.c
+++ b/ext/bson/write.c
@@ -34,6 +34,13 @@ static void pvt_put_bson_key(byte_buffer_t *b, VALUE string);
 static VALUE pvt_bson_byte_buffer_put_bson_partial_string(VALUE self, const char *str, int32_t length);
 static VALUE pvt_bson_byte_buffer_put_binary_string(VALUE self, const char *str, int32_t length);
 
+/* Raise ArgumentError if length exceeds the BSON int32_t string-length limit. */
+static void pvt_check_string_length(long length) {
+  if (length > INT32_MAX) {
+    rb_raise(rb_eArgError, "String length %ld exceeds BSON maximum", length);
+  }
+}
+
 static int fits_int32(int64_t i64){
   return i64 >= INT32_MIN && i64 <= INT32_MAX;
 }
@@ -211,7 +218,7 @@ static VALUE pvt_bson_encode_to_utf8(VALUE string) {
   VALUE encoding;
   VALUE utf8_string;
   const char *str;
-  int32_t length;
+  long length;
 
   existing_encoding_name = rb_funcall(
     rb_funcall(string, rb_intern("encoding"), 0),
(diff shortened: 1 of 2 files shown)


ext/bson/write.c
```diff
@@ -34,6 +34,13 @@ static void pvt_put_bson_key(byte_buffer_t *b, VALUE string);
 static VALUE pvt_bson_byte_buffer_put_bson_partial_string(VALUE self, const char *str, int32_t length);
 static VALUE pvt_bson_byte_buffer_put_binary_string(VALUE self, const char *str, int32_t length);
 
+/* Raise ArgumentError if length exceeds the BSON int32_t string-length limit. */
+static void pvt_check_string_length(long length) {
+  if (length > INT32_MAX) {
+    rb_raise(rb_eArgError, "String length %ld exceeds BSON maximum", length);
+  }
+}
```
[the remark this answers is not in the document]
jamis: Good catch. Fixed in 8f16834 — the bound is now `INT32_MAX - 5` rather than `INT32_MAX`, with a comment explaining that the binary-string path's `length + 5` arithmetic (4-byte int32 length prefix + payload + 1-byte null terminator) is what drives the tighter limit.

ext/bson/write.c
```diff
@@ -238,17 +250,18 @@ VALUE rb_bson_byte_buffer_put_string(VALUE self, VALUE string)
 {
   VALUE utf8_string;
   const char *str;
-  int32_t length;
+  long length;
```
comandeo-mongo, line 253: Why do we want to change this?

comandeo-mongo (approved):

ext/bson/write.c
jamis: It's because `RSTRING_LEN` returns `long`, not `int32_t`, so the result could (in extreme cases) be cast from `long` to a large negative `int32_t`, which could (potentially) cause problems.

